PR #3, opened 2017-04-27
Pjax integration

bdunogier:
> [EZP-27282](https://jira.ez.no/browse/EZP-27282)

Bridges the hybrid UI with the PJAX API in ezsystems/platform-ui-bundle.

It parses Responses from PJAX requests thanks to their semantic nature, and configures the App based on their contents. It maps the PJAX server-side-view content to the MainContent, and the PJAX title to the App's title.

capture d ecran 2017-05-12 a 13 21 23

### TODO
- [x] Issue link
- [x] Describe
- [x] Handle redirections (in general and after POST). Can be a follow-up, it works without it.
- [x] Handle navigation toggling for sub-routes (ex: sub parts of Section)
- [x] Fix phpspec
- [x] Squash most commits

### DIFF
--- a/src/lib/EventSubscriber/PjaxSubscriber.php
+++ b/src/lib/EventSubscriber/PjaxSubscriber.php
@@ -1,0 +1,76 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\HybridPlatformUi\EventSubscriber;
+
+use EzSystems\HybridPlatformUi\Components\App;
+use EzSystems\HybridPlatformUi\Mapper\MainContentMapper;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpKernel\Event\FilterResponseEvent;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\HttpKernel\KernelEvents;
+
+/**
+ * Catches admin PJAX requests, and maps them to Hybrid views.
+ */
+class PjaxSubscriber implements EventSubscriberInterface
+{
+    /**
+     * @var \Symfony\Component\HttpFoundation\RequestMatcherInterface
+     */
+    private $adminRequestMatcher;
+
+    /**
+     * @var \Symfony\Component\HttpFoundation\RequestMatcherInterface
+     */
+    private $pjaxRequestMatcher;
+
+    /**
+     * @var \EzSystems\HybridPlatformUi\Pjax\PjaxResponseMainContentMapper
+     */
+    private $responseMapper;
+
+    public static function getSubscribedEvents()
+    {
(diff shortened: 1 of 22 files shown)


src/lib/Pjax/PjaxResponseMainContentMapper.php
```diff
@@ -0,0 +1,86 @@
...
+            throw new \Exception(
+                "Error(s) occurred parsing the PJAX response:\n" .
+                implode("\n", $errors)
+            );
+        }
+
+        $xpath = new \DOMXPath($doc);
+        $nodeList = $xpath->query('//div[@data-name="title"]');
+        $title = $nodeList[0]->nodeValue;
+        $contentNodeList = $xpath->query('//div[@data-name="html"]//section[contains(concat(" ", normalize-space(@class), " "), " ez-serverside-content ")]');
```
dpobel: any reason to only keep `section.ez-serverside-content` ?

src/lib/Pjax/PjaxResponseMainContentMapper.php
```diff
@@ -0,0 +1,86 @@
...
+
+        $contentNode = $contentNodeList[0];
+        $attributes = 'ez-view-serversideview';
+        foreach (explode(' ', $contentNode->getAttribute('class')) as $classAttribute) {
+            if ($classAttribute !== 'ez-view-serversideview') {
+                $attributes .= " $classAttribute";
+            }
+        }
+        $content =
```
dpobel: depending on th anwser to my previous comment, I think `$content` could contain the content of the `div[data-name="html"]` which would greatly simplify the code there

src/lib/Pjax/PjaxResponseMainContentMapper.php
bdunogier: Not that I know of... it was a while ago :)

src/lib/Pjax/PjaxResponseMainContentMapper.php
bdunogier: I'll look into this, thank you.

src/lib/EventSubscriber/PjaxSubscriber.php
```diff
@@ -0,0 +1,86 @@
...
+            return;
+        }
+
+        $response = $event->getResponse();
+
+        // If AJAX update, follow the redirection and return the update for it.
+        // If not an AJAX update, send the redirection.
+        if ($response instanceof RedirectResponse) {
+            $event->setResponse(
+                new RedirectResponse($response->headers->get('PJAX-Location'))
```
dpobel: this kind of undoes the custom redirect mechanism defined by [FormProcessingDoneResponse](https://github.com/ezsystems/PlatformUIBundle/blob/2.0/Http/FormProcessingDoneResponse.php) in PlatformUIBundle. I think we should completely get rid of the FormProcessingDoneResponse (which was introduced because `XmlHttpRequest` does not allow to _track_ redirections but `window.fetch` used in `ez-platform-ui-app` has an option for that) and this can be removed as well.

bdunogier: https://github.com/ezsystems/PlatformUIBundle/pull/864

bdunogier:
All cleaned up, and sub-routes are now matched as well (e.g. sections will be toggled when editing a section).

bdunogier:
Ready for review.

dpobel (approved):

PR #6, opened 2017-05-19
EZP-27401: Add CSS to get the base colors and layout

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27401
A big part of it is in https://github.com/ezsystems/PlatformUIBundle/pull/867

# Description

Trying to apply the UI guidelines

[](http://www.youtube.com/watch?v=9FcN2ENw33I "Click to play on Youtube.com")

Found following issues in the guidelines:

* [Missing colors](https://github.com/ezsystems/ezsystems.github.com/issues/43)
* [Improve color names](https://github.com/ezsystems/ezsystems.github.com/issues/44)
* PR: https://github.com/ezsystems/ezsystems.github.com/pull/45 to keep Hybrid Platform UI button style and UI guidelines in synch


src/bundle/Resources/public/css/variables.css
```diff
@@ -0,0 +1,37 @@
+:root {
+    --ez-base-font-family: "Noto Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
+
+    /* text colors */
+    --ez-color-text-base: #333;
+    --ez-color-text-white: #fff; /* should probably not be called text white */
```
dpobel: following custom properties are based on the names in http://ezsystems.github.io/ui-guidelines/#resources-colors but I think some names should be improved:

* *Text white* resulting in `--ez-color-text-white` should not include the actual color, if at some point we change it to something, *white* won't be white any longer :) I would suggest `--ez-color-text-inverted`
* Names that includes a component or a part of the UI should be changed, most likely the same color will be used in others contexts (maybe `--ez-color-hyperlink` can be an exception to that rule given usually this color is unique). The name should rather be what it means (neutral, primary, secondary are good for instance). I know, naming is hard ;-)

ping @SylvainGuittard @inakijv

src/bundle/Resources/public/css/variables.css
```diff
@@ -0,0 +1,37 @@
+:root {
+    --ez-base-font-family: "Noto Sans", "Helvetica Neue", Helvetica, Arial, sans-serif;
```
dpobel: "Noto Sans" is not really a standard font (yet), I put what we used currently in PlatformUI 1.x as a fallback, any opinion on that @SylvainGuittard @inakijv ?

src/bundle/Resources/public/css/structure.css
```diff
@@ -0,0 +1,36 @@
...
+    flex: 0 0;
+}
+
+main {
+    flex: 1 0;
+    display: flex;
+}
+
+#discovery {
+    flex: 0 0 230px;
```
dpobel: @SylvainGuittard @inakijv
this is based on PlatformUI 1.x layout so I guess this will be different, especially with eZ action button having icons on top of the text, but maybe this should also be responsive ?

dpobel (commented):
@SylvainGuittard @inakijv also, we'll soon need some wireframe, at least to start structuring the app (details and colors can wait a bit)

src/bundle/Resources/public/css/button.css
```diff
@@ -0,0 +1,83 @@
...
+    white-space: nowrap;
+    vertical-align: middle;
+    text-align: center;
+    cursor: pointer;
+    -webkit-user-drag: none;
+    user-select: none;
+    box-sizing: border-box;
+
+    text-decoration: none;
+    padding: 0.5em 1em;
```
dpobel: in the styleguide, it's `padding: .85rem 1.75rem;` IMHO, that's really big

src/bundle/Resources/public/css/variables.css
SylvainGuittard: You're right `Noto` is not a standard font. The idea behind using Noto is to support a lot of languages for our new admin interface. We were thinking to include Noto as a font in the Hybride-Platform-UI bundle.
Regarding the fallback to Helvetica or Arial, we can keep it. +1

inakijv: I'd suggest `Arial, Helvetica, sans-serif` as the fall-back combination order

src/bundle/Resources/public/css/button.css
inakijv: In what part of the application are you implementing this? Do you have a screenshot?
`ez-button` is the default value, so there are specific classes, maybe we can specify this.

src/bundle/Resources/public/css/structure.css
inakijv: We are currently working with these properties in our testing environment, and we'll adjust based on your implementation

```  left: 0;
    display: inline-block;
    width: 115px;
    position: fixed;
    background-color: #565e63;
```

src/bundle/Resources/public/css/variables.css
dpobel: OK, created https://jira.ez.no/browse/EZP-27402 for that

src/bundle/Resources/public/css/button.css
dpobel: @inakijv `.ez-button` is global but I get why there's such a difference. In the UI guide lines, you use bootstrap.css which defines `html {font-size: 10px}` so in the end, 0.5em/1em here is almost the same of .85rem/1.75rem (1 or 2px or difference :))

src/bundle/Resources/public/css/structure.css
dpobel: OK, I'll adjust the width to 115px by default. For the rest (especially `position: fixed`), it really depends on the expected behavior while scrolling.

src/bundle/Resources/public/css/structure.css
```diff
@@ -0,0 +1,64 @@
...
+}
+
+#discovery {
+    flex: 0 0 115px;
+    width: 115px;
+    background: var(--ez-color-discovery-bar);
+    color: var(--ez-color-text-white);
+}
+
+ez-toolbar .ez-button {
```
dpobel: for buttons in discovery/action bar, I think it's just a special case for ez-button, so I think there's no need for `ez-action` and `ez-action-*` classes. Any opinion @inakijv @SylvainGuittard ?

Also, for icons, I think we should now use SVG instead of font icons (see https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/)

src/bundle/Resources/public/css/variables.css
dpobel: in addition, the action bar and discovery bar action button background colors and background color when hover are missing.

inakijv:
Wireframes requested for start structuring the app. Ping me if you need more than these ones :-)

content view view tab
content view details tab

src/bundle/Resources/public/css/structure.css
inakijv: No problem with that option :-) I treated them separately thinking on the extensibility cases these two bars have, highlighting they have to have a specific class given they are a particular case for buttons where they are grouped together and placed on both sides of the interface (plus the differences of style). But it's clear that from the code perspective they are what they are: buttons.

Ping us when you have more insights about the final option for icons :-)

src/bundle/Resources/public/css/variables.css
```diff
@@ -25,13 +27,13 @@
...
 
     /* background colors */
-    --ez-color-bg: #fafafa;
-    --ez-color-bg-header: #e7e7e7; /* maybe okish, but would prefer a color name */
-    --ez-color-bg-content: #f3f3f3; /* bad name, as it'll probably be used elsewhere */
-    --ez-color-bg-section-main: #e3e3e3;  /* bad name, as it'll probably be used elsewhere */
-    --ez-color-bg-label: #dbdbdb; /* bad name, as it'll probably be used elsewhere */
-    --ez-color-bg-fieldtype: #bbb; /* bad name, as it'll probably be used elsewhere */
+    --ez-color-ground-base: #fafafa;
+    --ez-color-ground-base-header: #e7e7e7;
```
inakijv: `--ez-color-ground-base-header: #e7e7e7;` wasn't included after the review & update. Where would it apply?

src/bundle/Resources/public/css/variables.css
```diff
@@ -25,13 +27,13 @@
...
+    --ez-color-ground-base-light: #f3f3f3;
+    --ez-color-ground-base-section-main: #e3e3e3;
+    --ez-color-ground-base-label: #dbdbdb;
```
inakijv, line 37: `--ez-color-ground-base-label` & `--ez-color-ground-base-fieldtype` Where would they apply?

src/bundle/Resources/public/css/variables.css
dpobel: forgot to remove some custom properties.
This is fixed in 13fd8ab9127a1e71ea1af05e64b7ca9e8f070406

PR #7, opened 2017-05-22
EZP-27356: Updating app struct to use pageTitle instead of title

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27356

## Description

Updating app component struct to use pageTitle app's property
see https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/10

### DIFF
--- a/src/lib/Components/App.php
+++ b/src/lib/Components/App.php
@@ -52,11 +52,6 @@ class App implements Component
         }
     }
 
-    public function getTitle()
-    {
-        return $this->title;
-    }
-
     protected function setToolbarsVisibility($config)
     {
         foreach ($this->toolbars as $toolbar) {
@@ -83,8 +78,8 @@ class App implements Component
         return [
             'selector' => self::TAG_NAME,
             'update' => [
-                'attributes' => [
-                    'title' => $this->title,
+                'properties' => [
+                    'pageTitle' => $this->title
                 ],
                 'children' => array_merge(
                     $this->toolbars,


src/lib/Components/App.php
```diff
@@ -52,9 +52,9 @@ public function setConfig(array $config)
         }
     }
 
-    public function getTitle()
```
dpobel: I think this method is not used anywhere, so you should just remove it

src/lib/Components/App.php
```diff
@@ -19,7 +19,7 @@ class App implements Component
 
     protected $templating;
 
-    protected $title;
+    protected $pageTitle;
```
dpobel: I would have kept this property like that. In this component, `title` is fine (it's really the app title)

dpobel (changes requested):

src/lib/Components/App.php
```diff
@@ -83,9 +83,6 @@ public function jsonSerialize()
         return [
             'selector' => self::TAG_NAME,
             'update' => [
-                'attributes' => [
```
dpobel: if you remove that, the title will never be updated (ok it's not completely implemented, see https://jira.ez.no/browse/EZP-27355).
So here, instead, you should update the property `pageTitle` by putting
```php
'update' => [
    'properties' => [
        'pageTitle' => $this->pageTitle
    ],
```

src/lib/Components/App.php
```diff
@@ -36,7 +36,7 @@ public function __construct(
     public function setConfig(array $config)
     {
         if (isset($config['title'])) {
-            $this->title = $config['title'];
+            $this->pageTitle = $config['title'];
```
dpobel: if you keep the `$title` property, you should use it everywhere :)

src/lib/Components/App.php
```diff
@@ -83,8 +78,8 @@ public function jsonSerialize()
         return [
             'selector' => self::TAG_NAME,
             'update' => [
-                'attributes' => [
-                    'title' => $this->title,
+                'properties' => [
+                    'pageTitle' => $this->pageTitle
```
dpobel: same as my previous comment :)

dpobel (changes requested):

dpobel (approved):

PR #9, opened 2017-05-29
EZP-27268: Remove navigation hub proto component

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27268

## Description

This PR is related to  https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/11 and removes the navigationhub proto components and adjust the markup for the 'real' navigationhub.


dpobel (changes requested):
In addition, after fixing comments in https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/11, you'll also need to adjust the markup generated by the NavigationHub component in https://github.com/ezsystems/hybrid-platform-ui/blob/master/src/bundle/Resources/views/components/navigationhub.html.twig#L13 so that both are consistent.

dpobel (approved):

PR #10, opened 2017-05-30
Custom main content element name

bdunogier:
> [EZP-27357](http://jira.ez.no/browse/EZP-27357)

Changes the markup so that the mainContent DOM element is defined by the MainContent template or result, and not the MainContent itself.

### TODO
- [x] Describe
- [x] Update the `PjaxMapper` so that it adds a root element (equivalent of locationview's `ez-content-view`)
- [x] ~~Change hybrid-ui-core-components string update~~ (https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/12)
- [x] See if `<ez-serverside-view>` is a good name for pjax pages
- [x] Fix conflicts
- [x] Update the CSS to match the markup

### DIFF
--- a/src/lib/Pjax/PjaxResponseMainContentMapper.php
+++ b/src/lib/Pjax/PjaxResponseMainContentMapper.php
@@ -54,7 +54,7 @@ class PjaxResponseMainContentMapper implements MainContentMapper
         $this->app->setConfig([
             'title' => $title,
             'toolbars' => ['discovery' => 1],
-            'mainContent' => ['result' => $content]
+            'mainContent' => ['result' => "<ez-serverside-view>" . $content . "</ez-serverside-view>"]
         ]);
     }
 
--- a/src/lib/Pjax/PjaxResponseMainContentMapper.php
+++ b/src/lib/Pjax/PjaxResponseMainContentMapper.php
@@ -54,7 +54,7 @@ class PjaxResponseMainContentMapper implements MainContentMapper
         $this->app->setConfig([
             'title' => $title,
             'toolbars' => ['discovery' => 1],
-            'mainContent' => ['result' => "<ez-serverside-view>" . $content . "</ez-serverside-view>"]
+            'mainContent' => ['result' => "<ez-serverside-content>" . $content . "</ez-serverside-content>"]
         ]);
     }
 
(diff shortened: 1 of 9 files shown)


dpobel:
HTML structure proposition:

```html
<ez-navigation-hub>

</ez-navigation-hub>
<div id="app-content">
    <ez-toolbar id="discovery">

    </ez-toolbar>
    <main>
        <ez-content-view>
            <div>
                    <h1>Content name</h1>
                </header>


                <ez-subitem parent-location-id=""></ez-subitem>
            </div>
            <ez-toolbar class="ez-action-bar" visible>

            </ez-toolbar>
        </ez-content-view>
    </main>
</div>
```

bdunogier:
Ping @dpobel about the newly added todo.

src/lib/Pjax/PjaxResponseMainContentMapper.php
```diff
@@ -54,7 +54,7 @@ public function map($response)
         $this->app->setConfig([
             'title' => $title,
             'toolbars' => ['discovery' => 1],
-            'mainContent' => ['result' => $content]
+            'mainContent' => ['result' => "<ez-serverside-view>" . $content . "</ez-serverside-view>"]
```
dpobel: on the tag name:
it is inspired by the PlatformUI `eZ.ServerSideView`, the view class used to handle pjax pages. But indeed, here that does not feel right. Maybe `<ez-server-side-content>` would be better ?

Also, since you start wrapping PJAX content with this tag, you should do the same for the Dashboard (dashboard.html.twig) and in locationview.html.twig where `<div class="ez-content-view">` becomes `<ez-server-side-content>` (until we need to specialize the behavior on those pages). With this last change, it would be nice
* to rename src/bundle/Resources/public/webcomponents/ez-main-content.html to src/bundle/Resources/public/webcomponents/ez-server-side-content.html (or which ever name we choose for the tag) and to update it to register the correct custom element
* and to update structure.css to not reference `ez-main-content`

dpobel (changes requested):

dpobel (approved):

bdunogier:
Works for me, @dpobel. Thank you :)

PR #11, opened 2017-06-01
EZP-27423: Browse should use locationId for UDW

StephaneDiot:
JIRA : https://jira.ez.no/browse/EZP-27423

## Description

Goes with https://github.com/ezsystems/PlatformUIBundle/pull/872 .
Makes the Browse to use locationId instead of REST id for the UDW to load the starting location

### DIFF
--- a/src/lib/Components/Browse.php
+++ b/src/lib/Components/Browse.php
@@ -9,19 +9,14 @@ use Symfony\Component\Routing\RouterInterface;
 
 class Browse implements Component
 {
-    protected $urlGenerator;
-
     /**
      * @var \Symfony\Component\HttpFoundation\Request
      */
     protected $request;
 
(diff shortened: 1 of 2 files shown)


src/lib/Components/Browse.php
```diff
@@ -32,26 +32,13 @@ protected function getLocationId()
         return "false";
     }
 
-    protected function getLocationRestId()
-    {
-        if ($this->request->attributes->get('location')) {
-            return $this->urlGenerator->generate(
```
dpobel: removing that makes the `urlGenerator` useless, so it should not be injected any more.

dpobel, line 38: together with that change, you should also remove the inline comment above the constructor

dpobel (changes requested):
in addition, you also need to adjust ez-browse.html

PR #15, opened 2017-06-12
EZP-27457: prevent authorization fatal errors in navigation hub

bdunogier:
> [EZP-27457](http://jira.ez.no/browse/EZP-27457)

Catches `UnauthorizedException` when generating links (Router -> UrlAliasRouter -> LoadLocation), and returns an empty link.

The empty link isn't a very good solution, but at least it avoids fatal errors. Should something be done on the frontend to complete this ? We also have the option of changing the link further from `navigationhub.html.twig`.

### DIFF
--- a/src/lib/NavigationHub/Link.php
+++ b/src/lib/NavigationHub/Link.php
@@ -11,4 +11,27 @@ abstract class Link
     public $name;
 
     abstract public function match(Request $request);
+
+    /**
+     * Returns the URL for the link.
+     *
+     * @return string|null The URL, or null if it could not be generated.
+     */
+    abstract public function getUrl();
+
+    public function __toString()
+    {
+        $disabled = '';
+        $url = $this->getUrl();
+
+        if ($url === null) {
+            $url = '#';
+            $disabled = ' data-disabled';
+        }
+
+        return sprintf(
+            '<a href="%s"%s>%s</a>',
+            $url, $disabled, htmlentities($this->name, ENT_HTML5)
+        );
+    }
 }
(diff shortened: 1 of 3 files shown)


dpobel:
yeah having an empty `href` attribute is not super nice. Also, if for such link, we want it to be visible in the interface but without effect, the link markup should be something like:
```html
<a href="#" data-disabled>Media</a>
```

(`href` can have any value, thanks to [base.css#L15](https://github.com/ezsystems/hybrid-platform-ui/blob/master/src/bundle/Resources/public/css/base.css#L15) a click on this link will have no effect)

So maybe as you suggest, we should change how links are generated. I would delegate the link markup generation to the Link object, ie in the navigation hub template we would have something like:

```twig
{% for link in links %}
<li data-zone="{{ link.zone }}" class="ez-zone-link {% if link.match(app.request) %} {{ matchedLinkClass }}{% endif %}">
   {{ link }}
</li>
{% endfor %}
```

And the Link object would be responsible for generating its markup and then it becomes easier to handle the UnauthorizedException exception. Also, this would allow integrators to change the generated markup for a given Link (can be useful if you want to add something richer than just a anchor).

bdunogier:
Changed to do what you've described, @dpobel.

src/lib/NavigationHub/Link.php
```diff
@@ -11,4 +11,27 @@
...
+        $url = $this->getUrl();
+
+        if ($url === null) {
+            $url = '#';
+            $disabled = ' data-disabled';
+        }
+
+        return sprintf(
+            '<a href="%s"%s>%s</a>',
+            $url, $disabled, htmlentities($this->name)
```
dpobel: nitpick: maybe add the ENT_HTML5 as a second parameter ? but maybe there's a SF service for that ?

dpobel (approved):

src/lib/NavigationHub/Link.php
bdunogier: Good catch, added it. I'd rather not use a service for this unless strictly necessary.

PR #18, opened 2017-06-15
EZP-27444 display details tab

mcurran-inviqa:
> [EZP-27444](https://jira.ez.no/browse/EZP-27444)

screenshot from 2017-06-15 17-28-02

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -9,6 +9,7 @@
 
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
@@ -33,6 +34,8 @@ class ContentViewController extends Controller
 
     public function detailsTabAction(ContentView $view)
     {
+        $repository = $this->getRepository();
+
         $versionInfo = $view->getContent()->getVersionInfo();
         $contentInfo = $versionInfo->getContentInfo();
 
@@ -40,8 +43,22 @@ class ContentViewController extends Controller
         $section = $sectionService->loadSection($contentInfo->sectionId);
 
         $userService = $this->getRepository()->getUserService();
-        $creator = $userService->loadUser($contentInfo->ownerId);
-        $lastContributor = $userService->loadUser($versionInfo->creatorId);
+
+        $creator = $repository->sudo(function () use ($contentInfo, $userService) {
+            try {
+                return $userService->loadUser($contentInfo->ownerId);
+            } catch (NotFoundException $e) {
+                return null;
+            }
+        });
+
+        $lastContributor = $repository->sudo(function () use ($versionInfo, $userService) {
+            try {
+                return $userService->loadUser($versionInfo->creatorId);
+            } catch (NotFoundException $e) {
+                return null;
+            }
+        });
 
         $location = $view->getLocation();
 
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -18,18 +18,18 @@ use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 class ContentViewController extends Controller
 {
     protected $defaultSortFields = [
-        Location::SORT_FIELD_NAME => 'sort.name',
-        Location::SORT_FIELD_PRIORITY => 'sort.priority',
-        Location::SORT_FIELD_MODIFIED => 'sort.modified',
-        Location::SORT_FIELD_PUBLISHED => 'sort.published'
+        Location::SORT_FIELD_NAME => ['key' => 'sort.name', 'default' => 'Content name'],
+        Location::SORT_FIELD_PRIORITY => ['key' => 'sort.priority', 'default' => 'Priority'],
+        Location::SORT_FIELD_MODIFIED => ['key' => 'sort.modified', 'default' => 'Modification date'],
+        Location::SORT_FIELD_PUBLISHED => ['key' => 'sort.published', 'default' => 'Publication date']
     ];
 
     protected $otherSortFields = [
-        Location::SORT_FIELD_PATH => 'sort.path',
-        Location::SORT_FIELD_CLASS_IDENTIFIER => 'sort.content.type.identifier',
-        Location::SORT_FIELD_SECTION => 'sort.section',
-        Location::SORT_FIELD_DEPTH => 'sort.depth',
-        Location::SORT_FIELD_CLASS_NAME => 'sort.content.type.name',
+        Location::SORT_FIELD_PATH => ['key' => 'sort.path', 'default' => 'Location path'],
+        Location::SORT_FIELD_CLASS_IDENTIFIER => ['key' => 'sort.content.type.identifier', 'default' => 'Content type identifier'],
+        Location::SORT_FIELD_SECTION => ['key' => 'sort.section', 'default' => 'Section'],
+        Location::SORT_FIELD_DEPTH => ['key' => 'sort.depth', 'default' => 'Location depth'],
+        Location::SORT_FIELD_CLASS_NAME => ['key' => 'sort.content.type.name', 'default' => 'Content type name']
     ];
 
     public function detailsTabAction(ContentView $view)
@@ -104,8 +104,8 @@ class ContentViewController extends Controller
     protected function getSortOrders()
     {
         return [
-            Location::SORT_ORDER_ASC => 'locationview.details.ascending',
-            Location::SORT_ORDER_DESC => 'locationview.details.descending'
+            Location::SORT_ORDER_ASC => ['key' => 'locationview.details.ascending', 'default' => 'Ascending'],
+            Location::SORT_ORDER_DESC => ['key' => 'locationview.details.descending', 'default' => 'Descending'],
         ];
     }
 }
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -1,0 +1,94 @@
+<?php
+
+/**
+ * File containing the ContentViewController class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\HybridPlatformUiBundle\Controller;
+
+use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+
+class ContentViewController extends Controller
+{
+    protected $defaultSortFields = [
+        Location::SORT_FIELD_NAME => 'sort.name',
+        Location::SORT_FIELD_PRIORITY => 'sort.priority',
+        Location::SORT_FIELD_MODIFIED => 'sort.modified',
+        Location::SORT_FIELD_PUBLISHED => 'sort.published'
+    ];
+
+    protected $otherSortFields = [
+        Location::SORT_FIELD_PATH => 'sort.path',
+        Location::SORT_FIELD_CLASS_IDENTIFIER => 'sort.content.type.identifier',
+        Location::SORT_FIELD_SECTION => 'sort.section',
+        Location::SORT_FIELD_DEPTH => 'sort.depth',
+        Location::SORT_FIELD_CLASS_NAME => 'sort.content.type.name',
+    ];
+
+    public function detailsTabAction(ContentView $view)
+    {
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        $sectionService = $this->getRepository()->getSectionService();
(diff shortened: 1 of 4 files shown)


src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -0,0 +1,89 @@
+<div class="ez-list-toolbar">
+    <h2 class="ez-list-toolbar-label">{{ 'locationview.details.title'|trans({}, 'locationview') }}</h2>
+</div>
+
+<table class="ez-table-data">
+    <thead>
+        <tr>
+            <th>{{ 'locationview.details.creator'|trans({}, 'locationview') }}</th>
+            <th>{{ 'locationview.details.lastcontributor'|trans({}, 'locationview') }}</th>
+            <th>{{ 'locationview.details.versions'|trans({}, 'locationview') }}</th>
```
mcurran-inviqa: note that this translation should be updated in i18n as the label is currently incorrect

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -0,0 +1,89 @@
...
+            <td>{{ section.id }}</td>
+        </tr>
+    </tbody>
+</table>
+
+<div class="ez-list-toolbar">
+    <h2 class="ez-list-toolbar-label">{{ 'locationview.details.subitems.default.ordering'|trans({}, 'locationview') }}</h2>
+</div>
+
+<p class="ez-subitem-ordering-details">{{ 'locationview.details.default.listing.of.subitems.by'|trans({}, 'locationview') }}
```
mcurran-inviqa: not mentioned in the ticket but replicated from v1 of the details tab as discussed in slack, this is currently using legacy markup, happy to updated if required

src/bundle/Resources/views/content/tabs/details.html.twig
bdunogier: We should try to use the same sources than in platform-ui v2. If possible, translators should not have to translate again (within reason).

src/bundle/Resources/config/views.yml
```diff
@@ -5,3 +5,8 @@ system:
                 default:
                     template: "EzSystemsHybridPlatformUiBundle::locationview.html.twig"
                     match: true
+            details:
```
bdunogier: I think we should to use something a bit more explanatory, like `details_tab` (and be consistent with subsequent tabs).

dpobel:
a link to the corresponding JIRA ticket, a screenshot and maybe some words in the description would be nice :)

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -0,0 +1,89 @@
...
+        <tr>
+            <th>{{ 'locationview.details.creator'|trans({}, 'locationview') }}</th>
+            <th>{{ 'locationview.details.lastcontributor'|trans({}, 'locationview') }}</th>
+            <th>{{ 'locationview.details.versions'|trans({}, 'locationview') }}</th>
+            <th>{{ 'locationview.details.translations'|trans({}, 'locationview') }}</th>
+        </tr>
+    </thead>
+    <tbody>
+        <tr>
+            <td>{{ ez_render_field(creator, 'first_name') }} {{ ez_render_field(creator, 'last_name') }} ({{ contentInfo.publishedDate|localizeddate('medium', 'short', app.request.locale) }})</td>
```
dpobel: you should not render the `first_name` and `last_name` fields as the User Content Type might be different and might not have those. Instead you should just put `{{ creator.name }}`
(also the line is quite long, for diff readability it's nice to wrap line to something like 120 characters)

src/bundle/Resources/views/content/tabs/details.html.twig
dpobel: hum if the string now has a different purpose, you should use a different key otherwise we'll have wrong translations

src/bundle/Resources/views/content/tabs/details.html.twig
dpobel: that's fine for now I guess. In any case, at some point I guess we'll have a Symfony Form, a Controller able to use it and we also need some JavaScript to automatically submit the form so in any case that's kind of broken :) Maybe to reflect that, for now the select tag should get a `disabled` attribute just to avoid wrong expectations

bdunogier (commented):
Looking good so far.

Could you please add a link to the JIRA story, as well as a short description and a screenshot of the result ?

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,94 @@
...
+            $sortFields[$currentSortField] = $this->otherSortFields[$currentSortField];
+        }
+
+        return $sortFields;
+    }
+
+    protected function getSortOrders()
+    {
+        return [
+            Location::SORT_ORDER_ASC => 'locationview.details.ascending',
```
dpobel: unsure about using translation keys like this. Won't this prevent us from generating translation file ? ping @bdunogier

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,94 @@
...
+    public function detailsTabAction(ContentView $view)
+    {
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        $sectionService = $this->getRepository()->getSectionService();
+        $section = $sectionService->loadSection($contentInfo->sectionId);
+
+        $userService = $this->getRepository()->getUserService();
+        $creator = $userService->loadUser($contentInfo->ownerId);
```
dpobel: the currently logged in user might not have the right to load those users, so this can throw an Exception. A possible improvement would be too use `Repository:sudo()` to still display user's name or to handle this Exception and for instance not displaying anything or a message or I don't know :)
@SylvainGuittard any opinion on that ?

dpobel (changes requested):

src/bundle/Resources/views/content/tabs/details.html.twig
mcurran-inviqa: purpose remains the same, but it was incorrect in v1 hence it should just be updated in the translation files which will fix for v1 also

src/bundle/Controller/ContentViewController.php
mcurran-inviqa: note we are using existing translations here, nothing new

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27444

screenshot from 2017-06-15 17-28-02

bdunogier:
Thank you for the screenshot. Moved the whole thing to the PR's description. Remember that it is an opensource product, and that not everybody knows what this is about exactly :)

src/bundle/Resources/views/content/tabs/details.html.twig
bdunogier: (yes, precisely 120 for github web reviews (too long for side by side, though ^^)

src/bundle/Controller/ContentViewController.php
bdunogier: Need to check how we handled those in v1, @dpobel.

There are tricks for some of them, @mcurran-inviqa. Not every string is as easy to extract. For now it will work, as the platform ui v1 strings are defined. But we need to make sure that the extractor knows about those somehow.

Note that I have created a story for i18n: https://jira.ez.no/browse/EZP-27494.

src/bundle/Controller/ContentViewController.php
SylvainGuittard: +1 for using `Repository:sudo()`.
We want to display the first name and the last name of the user account even if the current logged in user does not have the right (content/read).

src/bundle/Resources/views/content/tabs/details.html.twig
supriyabhargava: @bdunogier - we will have to incorporate the label change in content details table. 
'Versions' will now be 'Published version'

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -0,0 +1,89 @@
...
+            </td>
+        </tr>
+    </tbody>
+</table>
+
+<div class="ez-list-toolbar">
+    <h2 class="ez-list-toolbar-label">{{ 'locationview.details.technicaldetails'|trans({}, 'locationview') }}</h2>
+</div>
+
+<table class="ez-table-data">
```
supriyabhargava: @bdunogier - Label change - 'Id' should be replaced with 'ID' for all labels in Technical and Section details table

src/bundle/Resources/views/content/tabs/details.html.twig
dpobel: if the string is different in English, the purpose is not really the same, 'Versions' and 'Published version' does not mean the same thing at all. So for me, it's clear we should use a different key than the one we were using in v1 as the translation will need to be updated.

src/bundle/Controller/ContentViewController.php
dpobel: by the way, the User might also have been removed, so no matter if `Repository::sudo()` is used or not, an Exception can still be thrown in some cases, so this should be handled. I guess not displaying anything in that is fine for now (that was legacy's behavior) unless @SylvainGuittard you want us to display something special in that case.

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $view->addParameters([
+            'section' => $section,
+            'contentInfo' => $contentInfo,
+            'versionInfo' => $versionInfo,
+            'creator' => $creator,
+            'lastContributor' => $lastContributor,
+            'translations' => $this->getTranslations($versionInfo),
+            'ordering' => [
+                'sortFields' => $this->getSortFields($location->sortField),
```
bdunogier: You can use `$view->getLocation->sortField`, since you don't use `$location` anywhere else.

src/bundle/Resources/views/locationview.html.twig
```diff
@@ -30,7 +30,7 @@
                 {% endfor %}
             </div>
             <div class="ez-tabs-panel" id="ez-tab-details">
-                <p>Some details should be generated here</p>
+                {{ render(controller('ez_content:viewContent', {'contentId': contentId, 'viewType': 'details_tab'})) }}
```
bdunogier: `ez_content:viewContent` is deprecated, you must use `ez_content:viewAction`.
Also, you should pass the `locationId` as an extra argument, in order to know which location is being viewed.

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -0,0 +1,95 @@
...
+        <tr>
+            <td>{{ section.name }}</td>
+            <td>{{ section.identifier }}</td>
+            <td>{{ section.id }}</td>
+        </tr>
+    </tbody>
+</table>
+
+<div class="ez-list-toolbar">
+    <h2 class="ez-list-toolbar-label">{{ 'locationview.details.subitems.default.ordering'|trans({}, 'locationview')|desc('Sub-items default ordering') }}</h2>
```
bdunogier: It would be better to set the translation domain globally for the template (`{% trans_default_domain "locationview" %}`, and omit the parameter in all `|trans` calls.

bdunogier (commented):
I'm testing it locally, but I have added a few extra comments.

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,111 @@
...
+
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        $sectionService = $this->getRepository()->getSectionService();
+        $section = $sectionService->loadSection($contentInfo->sectionId);
+
+        $userService = $this->getRepository()->getUserService();
+
+        $creator = $repository->sudo(function () use ($contentInfo, $userService) {
```
bdunogier: We should add a `loadUser` protected method to the class, and use it to load the `$creator` and last contributor.

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,111 @@
...
+        });
+
+        $location = $view->getLocation();
+
+        $view->addParameters([
+            'section' => $section,
+            'contentInfo' => $contentInfo,
+            'versionInfo' => $versionInfo,
+            'creator' => $creator,
+            'lastContributor' => $lastContributor,
```
bdunogier: Once the `loadUser` method suggested above has been added, you can call it inline here.

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,111 @@
...
+            }
+        });
+
+        $location = $view->getLocation();
+
+        $view->addParameters([
+            'section' => $section,
+            'contentInfo' => $contentInfo,
+            'versionInfo' => $versionInfo,
+            'creator' => $creator,
```
bdunogier: Once the `loadUser` method suggested above has been added, you can call it inline here.

bdunogier (changes requested):

bdunogier (approved):

dpobel (approved):
:+1:

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,105 @@
...
+        ]);
+
+        return $view;
+    }
+
+    protected function loadUser($userId)
+    {
+        $userService = $this->getRepository()->getUserService();
+
+        return $this->getRepository()->sudo(function () use ($userId, $userService) {
```
andrerom, line 66: a bit late, but afaik you don't need to use sudo when using UserService, most load methods there don't check permissions on purpose for use cases where you need to load users _(REST however does check permissions when looking up users)_

PR #19, opened 2017-06-15
EZP-27445: Display versions tab information

maxbaldanza:
- This PR is for displaying the data only, editing will be done seperatly
- `locationview.versions.modified.language` Modified language does not exist in the translations yet so this will need adding for other languages

https://jira.ez.no/browse/EZP-27445
version_tab

### DIFF
--- a/spec/Filter/VersionFilterSpec.php
+++ b/spec/Filter/VersionFilterSpec.php
@@ -1,0 +1,70 @@
+<?php
+
+namespace spec\EzSystems\HybridPlatformUi\Filter;
+
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+use EzSystems\HybridPlatformUi\Filter\VersionFilter;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class VersionFilterSpec extends ObjectBehavior
+{
+    function it_is_initializable()
+    {
+        $this->shouldHaveType(VersionFilter::class);
+    }
+
+    function it_should_filter_drafts(
+        VersionInfo $versionInfo1,
+        VersionInfo $versionInfo2,
+        VersionInfo $versionInfo3
+    ) {
+        $versions = [
+            $versionInfo1,
+            $versionInfo2,
+            $versionInfo3
+        ];
+
+        $versionInfo1->isDraft()->willReturn(true)->shouldBeCalled();
+        $versionInfo2->isDraft()->willReturn(false)->shouldBeCalled();
+        $versionInfo3->isDraft()->willReturn(true)->shouldBeCalled();
+
+        $this->filterDrafts($versions)->shouldBeLike([$versionInfo1, $versionInfo3]);
+    }
+
+    function it_should_filter_published(
+        VersionInfo $versionInfo1,
+        VersionInfo $versionInfo2,
+        VersionInfo $versionInfo3
+    ) {
+        $versions = [
+            $versionInfo1,
+            $versionInfo2,
+            $versionInfo3
+        ];
+
+        $versionInfo1->isPublished()->willReturn(false)->shouldBeCalled();
+        $versionInfo2->isPublished()->willReturn(true)->shouldBeCalled();
+        $versionInfo3->isPublished()->willReturn(false)->shouldBeCalled();
+
+        $this->filterPublished($versions)->shouldBeLike([$versionInfo2]);
+    }
+
+    function it_should_filter_archived(
+        VersionInfo $versionInfo1,
+        VersionInfo $versionInfo2,
+        VersionInfo $versionInfo3
+    ) {
+        $versions = [
+            $versionInfo1,
+            $versionInfo2,
+            $versionInfo3
+        ];
+
+        $versionInfo1->isArchived()->willReturn(false)->shouldBeCalled();
+        $versionInfo2->isArchived()->willReturn(true)->shouldBeCalled();
+        $versionInfo3->isArchived()->willReturn(true)->shouldBeCalled();
+
+        $this->filterArchived($versions)->shouldBeLike([$versionInfo2, $versionInfo3]);
+    }
+}
--- a/spec/Filter/VersionFilterSpec.php
+++ b/spec/Filter/VersionFilterSpec.php
@@ -5,7 +5,6 @@ namespace spec\EzSystems\HybridPlatformUi\Filter;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use EzSystems\HybridPlatformUi\Filter\VersionFilter;
 use PhpSpec\ObjectBehavior;
-use Prophecy\Argument;
 
 class VersionFilterSpec extends ObjectBehavior
 {
@@ -22,7 +21,7 @@ class VersionFilterSpec extends ObjectBehavior
         $versions = [
             $versionInfo1,
             $versionInfo2,
-            $versionInfo3
+            $versionInfo3,
         ];
 
         $versionInfo1->isDraft()->willReturn(true)->shouldBeCalled();
@@ -40,7 +39,7 @@ class VersionFilterSpec extends ObjectBehavior
         $versions = [
             $versionInfo1,
             $versionInfo2,
-            $versionInfo3
+            $versionInfo3,
         ];
 
         $versionInfo1->isPublished()->willReturn(false)->shouldBeCalled();
@@ -58,7 +57,7 @@ class VersionFilterSpec extends ObjectBehavior
         $versions = [
             $versionInfo1,
             $versionInfo2,
-            $versionInfo3
+            $versionInfo3,
         ];
 
         $versionInfo1->isArchived()->willReturn(false)->shouldBeCalled();
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -14,6 +14,7 @@ use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
 use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+use EzSystems\HybridPlatformUi\Filter\VersionFilter;
 
 class ContentViewController extends Controller
 {
@@ -34,6 +35,16 @@ class ContentViewController extends Controller
         Location::SORT_FIELD_DEPTH => ['key' => 'sort.depth', 'default' => 'Location depth'],
         Location::SORT_FIELD_CLASS_NAME => ['key' => 'sort.content.type.name', 'default' => 'Content type name']
     ];
+
+    /**
+     * @var VersionFilter
+     */
+    private $versionFilter;
+
+    public function __construct(VersionFilter $versionFilter)
+    {
+        $this->versionFilter = $versionFilter;
+    }
 
     public function detailsTabAction(ContentView $view)
     {
@@ -65,17 +76,15 @@ class ContentViewController extends Controller
         $contentService = $this->getRepository()->getContentService();
         $versions = $contentService->loadVersions($contentInfo);
 
-        $versionFilter = $this->container->get('ezsystems.platform_ui.hybrid.filter.version_filter');
-
         $authors = [];
         foreach ($versions as $version) {
             $authors[$version->id] = $this->loadUser($version->creatorId);
         }
 
         $view->addParameters([
-            'draftVersions' => $versionFilter->filterDrafts($versions),
-            'publishedVersions' => $versionFilter->filterPublished($versions),
-            'archivedVersions' => $versionFilter->filterArchived($versions),
+            'draftVersions' => $this->versionFilter->filterDrafts($versions),
+            'publishedVersions' => $this->versionFilter->filterPublished($versions),
+            'archivedVersions' => $this->versionFilter->filterArchived($versions),
             'authors' => $authors
         ]);
 
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -81,11 +81,17 @@ class ContentViewController extends Controller
             $authors[$version->id] = $this->loadUser($version->creatorId);
         }
 
+        $translations = [];
+        foreach ($versions as $version) {
+            $translations[$version->id] = $this->getTranslations($version);
+        }
+
         $view->addParameters([
             'draftVersions' => $this->versionFilter->filterDrafts($versions),
             'publishedVersions' => $this->versionFilter->filterPublished($versions),
             'archivedVersions' => $this->versionFilter->filterArchived($versions),
-            'authors' => $authors
+            'authors' => $authors,
+            'translations' => $translations,
         ]);
 
         return $view;
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -9,6 +9,7 @@
 
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\API\Repository\Repository;
 use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 
@@ -16,15 +17,10 @@ class ContentViewController extends Controller
 {
     public function versionsTabAction(ContentView $view)
     {
+        $repository = $this->getRepository();
         $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
-        $contentService = $this->getRepository()->getContentService();
+        $contentService = $repository->getContentService();
         $versions = $contentService->loadVersions($contentInfo);
-
-        $userService = $this->getRepository()->getUserService();
-        $authors = [];
-        foreach ($versions as $version) {
-            $authors[$version->id] = $userService->loadUser($version->creatorId);
-        }
 
         $versionFilter = $this->container->get('ezsystems.platform_ui.hybrid.filter.version_filter');
 
@@ -32,9 +28,24 @@ class ContentViewController extends Controller
             'draftVersions' => $versionFilter->filterDrafts($versions),
             'publishedVersions' => $versionFilter->filterPublished($versions),
             'archivedVersions' => $versionFilter->filterArchived($versions),
-            'authors' => $authors
+            'authors' => $this->loadUsersForVersions($repository, $versions)
         ]);
 
         return $view;
+    }
+
+    private function loadUsersForVersions(Repository $repository, array $versions)
+    {
+        $authors = [];
+        $userService = $repository->getUserService();
+        foreach ($versions as $version) {
+            $authors[$version->id] = $repository->sudo(
+                function () use ($version, $userService) {
+                    return $userService->loadUser($version->creatorId);
+                }
+            );
+        }
+
+        return $authors;
     }
 }
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * File containing the ContentViewController class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\HybridPlatformUiBundle\Controller;
+
+use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+
+class ContentViewController extends Controller
+{
+    public function versionsTabAction(ContentView $view)
+    {
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $contentService = $this->getRepository()->getContentService();
+        $versions = $contentService->loadVersions($contentInfo);
+
+        $userService = $this->getRepository()->getUserService();
+        $authors = [];
+        foreach ($versions as $version) {
+            $authors[$version->id] = $userService->loadUser($version->creatorId);
+        }
+
+        $versionFilter = $this->container->get('ezsystems.platform_ui.hybrid.filter.version_filter');
+
+        $view->addParameters([
+            'draftVersions' => $versionFilter->filterDrafts($versions),
+            'publishedVersions' => $versionFilter->filterPublished($versions),
+            'archivedVersions' => $versionFilter->filterArchived($versions),
+            'authors' => $authors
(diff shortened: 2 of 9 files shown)


src/bundle/Resources/views/content/tabs/versions/table.html.twig
```diff
@@ -0,0 +1,32 @@
...
+    {% for version in versions %}
+        <tr>
+            <td>
+                {{ version.versionNo}}
+            </td>
+            <td>
+                {{ version.getContentInfo().mainLanguageCode }}
+            </td>
+            <td>
+                {{ ez_render_field(authors[version.id], 'first_name') }} {{ ez_render_field(authors[version.id], 'last_name') }}
```
dpobel: same remark as in https://github.com/ezsystems/hybrid-platform-ui/pull/18#discussion_r122197912, `first_name` and `last_name` fields might not be there, `{{ authors[version.id].name }}` should display the user full name

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,40 @@
...
+    public function versionsTabAction(ContentView $view)
+    {
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $contentService = $this->getRepository()->getContentService();
+        $versions = $contentService->loadVersions($contentInfo);
+
+        $userService = $this->getRepository()->getUserService();
+        $authors = [];
+        foreach ($versions as $version) {
+            $authors[$version->id] = $userService->loadUser($version->creatorId);
```
dpobel: here as well, same remark as inhttps://github.com/ezsystems/hybrid-platform-ui/pull/18#discussion_r122208466 (so ping @SylvainGuittard to define the expected behavior)

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,18 @@
+<div class="ez-versions-list-container">
```
dpobel: I think you can remove this `div`

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,18 @@
...
+    {%  if draftVersions %}
+        <h2 class="ez-list-toolbar-label">{{'locationview.versions.draft.under.edit'|trans({}, 'locationview')}}</h2>
```
dpobel: this `h2` should be inside a `div` with the class `ez-list-toolbar` (valid also for following `h2` as well)

src/bundle/Resources/views/content/tabs/versions/table.html.twig
```diff
@@ -0,0 +1,32 @@
...
+        </tr>
+    </thead>
+    <tbody>
+    {% for version in versions %}
+        <tr>
+            <td>
+                {{ version.versionNo}}
+            </td>
+            <td>
+                {{ version.getContentInfo().mainLanguageCode }}
```
dpobel: I would write it `version.contentInfo.mainLanguageCode`, it feels more *template* :)

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Filter;
+
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+
+class VersionFilter
+{
+    public function filterDrafts(array $versions)
```
dpobel: those public methods and the class deserve to have some documentation (also the file is missing the standard header)

dpobel (changes requested):
as for the other PR, a screenshot would be nice :)

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,18 @@
...
+
+    {% endif %}
+
+    {%  if publishedVersions %}
+        <h2 class="ez-list-toolbar-label">{{'locationview.versions.published.version'|trans({}, 'locationview')}}</h2>
+        {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': publishedVersions} %}
+    {% endif %}
+
+    {%  if archivedVersions %}
+        <h2 class="ez-list-toolbar-label">{{'locationview.versions.archived.versions'|trans({}, 'locationview')}}</h2>
```
maxbaldanza: I think this translation should be `locationview.versions.archived.version` to keep them consistent. Could this safely be changed in the translations repo?
https://github.com/ezsystems/ezplatform-i18n

bdunogier: ezplatform-i18n is read-only, it gets the sources from each individual package.
We must start adding translation sources to this repository, and sync the files to ezplatform-i18n.

About translation strings, we use JmsTranslationBundle, that provides |desc twig function. It is used to provide the default translation while using a static identifier. It would be good to add the translations straight away.

Example:
```
{{'locationview.versions.archived.versions'|trans({}, 'locationview')|desc('Archived versions')}}
```

You can find examples in the [kernel](https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Bundle/EzPublishCoreBundle/Resources/views/fielddefinition_settings.html.twig#L18).

I'll create a story to configure translations extraction in this repository.

spec/Filter/VersionFilterSpec.php
```diff
@@ -0,0 +1,70 @@
+<?php
+
+namespace spec\EzSystems\HybridPlatformUi\Filter;
+
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+use EzSystems\HybridPlatformUi\Filter\VersionFilter;
+use PhpSpec\ObjectBehavior;
+use Prophecy\Argument;
+
+class VersionFilterSpec extends ObjectBehavior
```
bdunogier: Love this 👍

src/bundle/Controller/ContentViewController.php
SylvainGuittard: +1 for using `Repository:sudo()`.
We want to display the first name and the last name of the user account even if the current logged in user does not have the right (content/read).

src/lib/Filter/VersionFilter.php
maxbaldanza: Good point, I was looking at some of the other classes in this library and they are missing too which is why it was missed. For example `EzSystems\HybridPlatformUi\EventSubscriber\AppRendererSubscriber`
`EzSystems\HybridPlatformUi\EventSubscriber\ComponentRendererSubscriber`
I shall add it in this class and add some documentation for the methods 😄

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,53 @@
...
+    public function versionsTabAction(ContentView $view)
+    {
+        $repository = $this->getRepository();
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $contentService = $repository->getContentService();
+        $versions = $contentService->loadVersions($contentInfo);
+
+        $authors = $this->loadUsersForVersions($repository, $versions);
+
+        $versionFilter = $this->container->get('ezsystems.platform_ui.hybrid.filter.verion_filter');
```
bdunogier: Typo in `version_filter`.

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,53 @@
...
+class ContentViewController extends Controller
+{
+    public function versionsTabAction(ContentView $view)
+    {
+        $repository = $this->getRepository();
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $contentService = $repository->getContentService();
+        $versions = $contentService->loadVersions($contentInfo);
+
+        $authors = $this->loadUsersForVersions($repository, $versions);
```
bdunogier: I don't think you need to create a local variable for this. You could move the call directly to the `addParameters` call below.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,22 @@
+{%  if draftVersions %}
```
bdunogier: CS: extra space after `%`.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,22 @@
...
+    <div class="ez-list-toolbar">
+        <h2 class="ez-list-toolbar-label">{{'locationview.versions.draft.under.edit'|trans({}, 'locationview')}}</h2>
+    </div>
+    {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': draftVersions} %}
+
+{% endif %}
+
+{%  if publishedVersions %}
```
bdunogier: CS: extra space after `%`.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,22 @@
...
+    <div class="ez-list-toolbar">
+        <h2 class="ez-list-toolbar-label">{{'locationview.versions.published.version'|trans({}, 'locationview')}}</h2>
+    </div>
+    {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': publishedVersions} %}
+{% endif %}
+
+{%  if archivedVersions %}
```
bdunogier: CS: extra space after `%`.

src/bundle/Resources/views/content/tabs/versions/table.html.twig
```diff
@@ -0,0 +1,32 @@
+<table class="ez-table-data">
+    <thead>
+        <tr>
+            <th>{{ 'locationview.versions.version'|trans({}, 'locationview') }}</th>
```
bdunogier: Missing `desc()` call:

```
{{ 'locationview.versions.version'|trans({}, 'locationview')|desc('Version') }
```

Same for the other `|trans()` calls.

bdunogier: You could also set the translation domain globally for the template. It would avoid having to repeat it in every `|trans()` call, thus skipping the empty hash as the first argument.

src/bundle/Resources/config/views.yml
```diff
@@ -5,3 +5,8 @@ system:
                 default:
                     template: "EzSystemsHybridPlatformUiBundle::locationview.html.twig"
                     match: true
+            versions:
```
bdunogier: We should have a convention for tab views. I suggest locationview_XXX_tab. Example: `locationview_versions_tab`. Naming is debatable (ping @dpobel).

bdunogier (changes requested):

bdunogier:
Also, given that the PR requires ezsystems/ezpublish-kernel#2017, merged into `master`, we need to update the ezpublish-kernel requirements. It should requires `^6.11` (the branch alias for the current master).

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    private function loadUsersForVersions(Repository $repository, array $versions)
+    {
+        $authors = [];
+        $userService = $repository->getUserService();
+        foreach ($versions as $version) {
+            $authors[$version->id] = $repository->sudo(
+                function () use ($version, $userService) {
+                    return $userService->loadUser($version->creatorId);
```
dpobel: the user might have been deleted so you have to handle a potential NotFoundException
(see for instance https://github.com/ezsystems/hybrid-platform-ui/pull/18/files#diff-3fdb78eacffbdaab16a64af286dfdc07R66)

also I hope the SPI cache works well on user loading because it's very likely the same user will be loaded several times (ping @andrerom)
If not, maybe you could index authors by user id and change the template accordingly ?

dpobel (changes requested):

src/bundle/Controller/ContentViewController.php
maxbaldanza: Ideally I'd like to use that method rather than re-creating it in this PR. I shall wait for that PR to be merged and then refactor to use that instead.

From what I've seen there looks to be cache layer in front of loading the users but I shall wait to see what Andre thinks.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,24 @@
+{% trans_default_domain "locationview" %}
+
+{% if draftVersions %}
+    <div class="ez-list-toolbar">
+        <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.draft.under.edit'|trans()|desc('Draft under edit') }}</h2>
+    </div>
+    {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': draftVersions} %}
```
bdunogier: According to knowledgeable Symfony users, `{{ include() }}` should be used instead of `{% include %}`: https://twig.sensiolabs.org/doc/2.x/functions/include.html.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,24 @@
...
+
+{% endif %}
+
+{% if publishedVersions %}
+    <div class="ez-list-toolbar">
+        <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.published.version'|trans()|desc('Published version') }}</h2>
+    </div>
+    {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': publishedVersions} %}
```
bdunogier: Include again.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -0,0 +1,24 @@
...
+{% endif %}
+
+{% if archivedVersions %}
+    <div class="ez-list-toolbar">
+        <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.archived.versions'|trans()|desc('Archived versions') }}</h2>
+    </div>
+    {% include '@EzSystemsHybridPlatformUi/content/tabs/versions/table.html.twig' with {'versions': archivedVersions} %}
```
bdunogier: And another include.

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+namespace EzSystems\HybridPlatformUi\Filter;
+
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+
+class VersionFilter
+{
+    /**
+     * Filters a list of versions to retrieve only the draft versions
```
bdunogier: CS: period at the end of the sentence.

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param array $versions
```
bdunogier: `@param eZ\Publish\API\Repository\Values\Content\VersionInfo[] $versions`

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @return array
```
bdunogier: `@return eZ\Publish\API\Repository\Values\Content\VersionInfo[]`

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function filterDrafts(array $versions)
+    {
+        return array_values(array_filter($versions, function (VersionInfo $version) {
+            return $version->isDraft();
+        }));
+    }
+
+    /**
+     * Filters a list of versions to retrieve only the published versions
```
bdunogier: CS: period at the end of the sentence.

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param array $versions
```
bdunogier: `@param eZ\Publish\API\Repository\Values\Content\VersionInfo[] $versions`

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @return array
```
bdunogier: `@return eZ\Publish\API\Repository\Values\Content\VersionInfo[]`

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     */
+    public function filterPublished(array $versions)
+    {
+        return array_values(array_filter($versions, function (VersionInfo $version) {
+            return $version->isPublished();
+        }));
+    }
+
+    /**
+     * Filters a list of versions to retrieve only the archived versions
```
bdunogier: CS: period at the end of the sentence.

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @param array $versions
```
bdunogier: `@param eZ\Publish\API\Repository\Values\Content\VersionInfo[] $versions`

src/lib/Filter/VersionFilter.php
```diff
@@ -0,0 +1,54 @@
...
+     *
+     * @return array
```
bdunogier: `@return eZ\Publish\API\Repository\Values\Content\VersionInfo[]`

src/bundle/Controller/ContentViewController.php
```diff
@@ -59,6 +59,29 @@ public function detailsTabAction(ContentView $view)
         return $view;
     }
 
+    public function versionsTabAction(ContentView $view)
+    {
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $contentService = $this->getRepository()->getContentService();
+        $versions = $contentService->loadVersions($contentInfo);
+
+        $versionFilter = $this->container->get('ezsystems.platform_ui.hybrid.filter.version_filter');
```
bdunogier: I'd prefer if we injected this somehow. I'll test some things with auto-wiring and I'll let you know.

bdunogier (changes requested):

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: Thanks, it's been a while since using twig and after a quick look it does look like the function is the one to use 👍

src/bundle/Controller/ContentViewController.php
maxbaldanza: I prefer services to be injected in too. We could make the controller a service that way it can be injected in?
I assumed getting things directly from the container was the convention as that's how it's done elsewhere: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/MVC/Symfony/Controller/Controller.php#L98

bdunogier: It's done like that in one controller, but several others do not. Good confirmation bias example ;)

Jokes aside, we may wanna try to auto-wiring of controller action arguments, and have the container pass the Filter service to the action method instead: https://symfony.com/doc/current/controller.html#controller-accessing-services.

Could you give it a try @maxbaldanza ?

andrerom:
Hmm, it's a nitpick so I don't want to comment on it inline. But looking at `VersionFilter` I'm unsure if we ended up with something better then what we had with status constants, cause with those it would have just been one method taking status as second argument somewhat cleanly using constant to refer to the one you need. @maxbaldanza Where do we make the cut between code duplication and code ascetics?

maxbaldanza:
@andrerom I've updated `VersionFilter` to remove some of the duplication if you want to take another look and see what your thoughts are? 
I would always suggest to use methods rather than directly checking against constants in lots of different places just so that we can encapsulate that logic and hold it all in one place in case there's a requirement in the future that would force it to change.

bdunogier:
Also, methods such as `isDraft()` allow tests such as `{% if content.draft %}`. Constants make it more complicated from an integrator's perspective.

bdunogier:
But thinking about it, I do see @andrerom's point. We define three methods in order to use:

```php
$versionFilter->filterDrafts($versions)
```

instead of 
```php
$versionFilter->filterByStatus($versions, VersionInfo::STATUS_DRAFT);
```

This part of the API is by essence _very_ stable.

I kinda feel like this introduces both a little duplication (not very severe, though), _and_ a coupling with which version statuses exist in the repository.

The point about being future-proof by abstracting this for the UI is a valid point, though. Even if the repository's API stays stable, the API's features might follow a different path. It also leverages the `isStatus()` methods from `VersionInfo`. If we have a method that uses the status as an argument, we can't use them.

In any case, let's not forget that this is far from being tagged. We can also see over time how things evolve in this area, and adapt accordingly.

andrerom:
@bdunogier you summarized up my view nicely, I was not saying the api additions where pointless at all, in cases like twig it simplifies the code a lot, it was specifically for a filter function where constant use seemed like a better fit.

bdunogier:
@maxbaldanza my PR that removes PHP 5.6 testing has been merged to master, please rebase to get rid of the travis failure.

bdunogier (dismissed):

src/bundle/Resources/config/views.yml
```diff
@@ -10,3 +10,8 @@ system:
                     template: "EzSystemsHybridPlatformUiBundle:content/tabs:details.html.twig"
                     controller: "EzSystemsHybridPlatformUiBundle:ContentView:detailsTab"
                     match: true
+            versions_tab:
+                default:
+                    template: "EzSystemsHybridPlatformUiBundle:content/tabs:versions.html.twig"
+                    controller: "EzSystemsHybridPlatformUiBundle:ContentView:versionsTab"
```
bdunogier, line 16: On my instance, this change fixes the autowiring of the controller action arguments:

```
diff --git a/src/bundle/Resources/config/views.yml b/src/bundle/Resources/config/views.yml
index 5647921..12a09c5 100644
--- a/src/bundle/Resources/config/views.yml
+++ b/src/bundle/Resources/config/views.yml
@@ -13,5 +13,5 @@ system:
             versions_tab:
                 default:
                     template: "EzSystemsHybridPlatformUiBundle:content/tabs:versions.html.twig"
-                    controller: "EzSystemsHybridPlatformUiBundle:ContentView:versionsTab"
+                    controller: 'EzSystems\HybridPlatformUiBundle\Controller\ContentViewController:versionsTabAction'
                     match: true
```

According to the Sf doc, if the controller uses its class as its ID, the `Bundle:Controller:Action` syntax can be used. But it doesn't seem to behave as expected. The argument isn't filled in by the ArgumentResolver.

bdunogier (approved):
But let's merge this without auto-wiring, we can take care of it later.

dpobel (approved):

PR #20, opened 2017-06-16
Setup i18n for the package (EZP-27494)

bdunogier:
> [EZP-27494](http://jira.ez.no/browse/EZP-27494)

Adds the translation extraction script as well as initial translations.
To extract the translations, run `sh bin/extract-translations.sh` from `vendor/ezsystems/hybrid-platform-ui`. The structure is important, as the script will move up to the ezplatform root directory to run `bin/console`.

### DIFF
--- a/src/bundle/Resources/translations/fieldview.en.xlf
+++ b/src/bundle/Resources/translations/fieldview.en.xlf
@@ -1,0 +1,17 @@
+<?xml version="1.0" encoding="utf-8"?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
+  <file date="2017-06-23T10:41:58Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+    <header>
+      <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
+      <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
+    </header>
+    <body>
+      <trans-unit id="977cb55c24b1708d46ccbd753c2cea745976ae21" resname="fieldview.field.empty">
+        <source>fieldview.field.empty</source>
+        <target>fieldview.field.empty</target>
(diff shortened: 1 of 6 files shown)


src/bundle/Resources/translations/fieldview.en.xlf
```diff
@@ -0,0 +1,17 @@
...
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
+  <file date="2017-06-15T17:29:53Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+    <header>
+      <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
+      <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
+    </header>
+    <body>
+      <trans-unit id="977cb55c24b1708d46ccbd753c2cea745976ae21" resname="fieldview.field.empty">
+        <source>fieldview.field.empty</source>
+        <target state="new">fieldview.field.empty</target>
```
dpobel: hum this string is already defined (and translated) in https://github.com/ezsystems/PlatformUIBundle/blob/master/Resources/translations/fieldview.en.xlf#L21 but of course the current extraction script does not know that which means (in the current state) we'll duplicate existing strings. So I guess we have to think a bit about how we organize/reuse translatable strings in various packages

bdunogier: Unless I'm mistaken, the translation system will load the translations from PlatformUI, since the domain is the same. hybrid-platform-ui defines its own sources, that's all (since it has those as source string as well).

Or am I missing something ?

dpobel: and in crowdin, won't those strings appear as "to be translated" ?

bdunogier: No, they should not. Probably not.

bdunogier: I'm not sure how crowdin will deal with the same string having multiple sources, but the same domain. I think we should merge and see :-)

PR #21, opened 2017-06-16
EZP-27501: use pjax as a pagelayout

bdunogier:
> Implements [EZP-27501](http://jira.ez.no/browse/EZP-27501)
> Status: ready for review

Configures `src/bundle/Resources/views/pagelayout.html.twig` as the global pagelayout for the `admin_group` siteaccess group. Core templates that use it, `such as content_edit.html.twig, will be rendered within the PJAX DOM and integrated in the Hybrid UI.

capture d ecran 2017-06-17 a 02 11 09

In addition, adds a `PjaxResponseMatcher`, used to detect PJAX responses by looking for `<div data-name="html">` in the content. It allows to match requests that don't identify by the URI or by the PJAX header.

capture d ecran 2017-06-17 a 02 11 59

### TODO
- [x] Find a better name for the pagelayout template
- [x] Add specs
- [x] Bonus: set the login pagelayout to pjax

### DIFF
--- a/src/bundle/Resources/views/pjax_layout.html.twig
+++ b/src/bundle/Resources/views/pjax_layout.html.twig
@@ -1,0 +1,15 @@
+<div data-name="title">{% block title %}{{ title|default('') }}{% endblock %}</div>
+
+<div data-name="html">
+    {% block content %}{% endblock %}
+</div>
+
+<ul data-name="notification">
+    {% block notification %}
+        {% set notifications = app.session.flashBag.get("notification") %}
+        {% for notification in notifications %}
+            <li data-state="{{ notification.state }}">{{ notification.message }}</li>
+
+        {% endfor %}
+    {% endblock %}
+</ul>
--- a/src/lib/Pjax/PjaxResponseMatcher.php
+++ b/src/lib/Pjax/PjaxResponseMatcher.php
@@ -1,0 +1,18 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Pjax;
+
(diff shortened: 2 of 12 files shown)


src/lib/Pjax/PjaxResponseMatcher.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+namespace EzSystems\HybridPlatformUi\Pjax;
+
+use EzSystems\HybridPlatformUi\Http\ResponseMatcherInterface;
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Matches a Pjax Http Response.
+ */
+interface PjaxResponseMatcher extends ResponseMatcherInterface
```
glye, line 15: Why is one named -Interface and the other not?

glye (approved):
FWIW

src/lib/Pjax/PjaxResponseMatcher.php
bdunogier: Yeah... 'cause I wanted `ResponseMatcherInterface` to be consistent with Symfony's `RequestMatcherInterface`. Do you see my dilemma ?

glye: Let's follow the trend, and go for Facey McInterface-face.

src/bundle/Resources/views/pjax_layout.html.twig
```diff
@@ -0,0 +1,15 @@
+<div data-name="title">{% block title %}{{ title|default('hybrid pagelayout title') }}{% endblock %}</div>
```
dpobel: the default title should probably be an empty string

dpobel (approved):
+1 besides the default title comment.
not strictly part of this story, we should check why we have a blank page after login or when providing wrong credentials.

src/bundle/Resources/views/pjax_layout.html.twig
bdunogier: Changed.

PR #22, opened 2017-06-19
EZP-27446 locations tab

mcurran-inviqa:
> https://jira.ez.no/browse/EZP-27446

image


mcurran-inviqa:
https://jira.ez.no/browse/EZP-27446
image

src/bundle/Controller/ContentViewController.php
```diff
@@ -0,0 +1,43 @@
...
+
+class ContentViewController extends Controller
+{
+    public function locationsTabAction(ContentView $view)
+    {
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        if ($contentInfo->published) {
+            $locationRepository = $this->getRepository()->getLocationService();
```
bdunogier: We call it `locationService`, not `locationRepository`.

src/bundle/Resources/views/fields/location/path.html.twig
```diff
@@ -0,0 +1,6 @@
+{% for location in locations %}
```
bdunogier: What about wrapping the whole thing with a `{spaceless}` tag ?

src/bundle/Resources/views/locationview.html.twig
```diff
@@ -36,7 +36,7 @@
                 <p>Versions list should be generated here</p>
             </div>
             <div class="ez-tabs-panel" id="ez-tab-locations">
-                <p>Locations list should be generated here</p>
+                {{ render(controller('ez_content:viewAction', {'contentId': contentId, 'locationID': locationId, 'viewType': 'locations_tab'})) }}
```
bdunogier: `'locationId'` (lowercase 'd') ?

src/lib/Decorator/LocationDecorator.php
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\HybridPlatformUi\Decorator;
+
+use eZ\Publish\API\Repository\Values\Content\Location;
+
+class LocationDecorator extends ValueObjectDecorator
```
bdunogier: Unless I'm mistaken, the methods in this file are redefined as forwarding the call to the decorated object because the magic methods in the abstract class are not working ? If yes, then these really need to be documented.

But does it overall mean that only the methods defined in this class will actually work (e.g. no magic methods ?

src/lib/Decorator/LocationDecorator.php
```diff
@@ -0,0 +1,51 @@
...
+{
+    /**
+     * @var int
```
bdunogier: The doc should specify that the value is meant to be calculated externally.

src/bundle/Controller/ContentViewController.php
```diff
@@ -59,6 +60,31 @@ public function detailsTabAction(ContentView $view)
...
     }
 
+    public function locationsTabAction(ContentView $view)
+    {
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        if ($contentInfo->published) {
+            $locationRepository = $this->getRepository()->getLocationService();
+            $locations = $locationRepository->loadLocations($contentInfo);
```
bdunogier: For readability, I'd refactor the loading of locations to a `loadLocations` private method (or an external service if you want to, it would make it testable).

src/bundle/Controller/ContentViewController.php
```diff
@@ -59,6 +60,31 @@ public function detailsTabAction(ContentView $view)
...
+
+            $locations = array_map(function (Location $location) {
+                return new LocationDecorator($location);
+            }, $locations);
+
+            foreach ($locations as $location) {
+                $location->childCount = $locationRepository->getLocationChildCount($location->getValueObject());
```
bdunogier: `childCount` can be set in the callback above, after creating the `LocationDecorator` object.

src/lib/Templating/Twig/LocationExtension.php
```diff
@@ -0,0 +1,66 @@
...
+
+        /** remove root location from display */
+        array_shift($path);
+        /** last element will be the provided location, no need to retrieve it again */
+        array_pop($path);
+
+        $locations = [];
+
+        foreach ($path as $locationId) {
+            $locations[] = $this->locationService->loadLocation($locationId);
```
bdunogier: I think we can, and should, do better than that. The `SearchService`, with `findLocations()`, can probably be used to load all locations at once. Also, I'd really prefer if the loading of content wasn't done using a twig function.

bdunogier (changes requested):

src/lib/Decorator/LocationDecorator.php
mcurran-inviqa: It's more because the likes of twig do some level of introspection of the code, these 3 methods are defined in the location value object.

the idea here is to implement all public methods and pass them to the original object

src/lib/Repository/DecoratedLocationRepository.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\Location;
+use EzSystems\HybridPlatformUi\Decorator\LocationDecorator;
+
+class DecoratedLocationRepository
```
bdunogier: DecoratedLocationService, not Repository. Repository has another meaning in the eZ Platform domain. Please also update the service name.

bdunogier:
Hey, it's working ! :)

spec/Repository/DecoratedLocationServiceSpec.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace spec\EzSystems\HybridPlatformUi\Repository;
```
bdunogier: Missing copyright header.

spec/Repository/PathServiceSpec.php
```diff
@@ -0,0 +1,39 @@
+<?php
+
+namespace spec\EzSystems\HybridPlatformUi\Repository;
```
bdunogier: Missing copyright header.

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,48 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Repository;
```
bdunogier: Missing copyright header.

src/lib/Repository/PathService.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Repository;
```
bdunogier: Missing copyright header.

src/lib/Repository/PathService.php
```diff
@@ -0,0 +1,33 @@
...
+
+use eZ\Publish\API\Repository\SearchService;
+use eZ\Publish\API\Repository\Values\Content\LocationQuery;
+use eZ\Publish\API\Repository\Values\Content\Query\Criterion\Ancestor;
+use eZ\Publish\API\Repository\Values\Content\Search\SearchHit;
+use eZ\Publish\Core\Repository\Values\Content\Location;
+
+class PathService
```
bdunogier: PhpDoc please.

src/lib/Repository/PathService.php
```diff
@@ -0,0 +1,33 @@
...
+     * @var SearchService
+     */
+    private $searchService;
+
+    public function __construct(SearchService $searchService)
+    {
+        $this->searchService = $searchService;
+    }
+
+    public function loadPathLocations(Location $location)
```
bdunogier: PhpDoc.

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,48 @@
...
+
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\Content\Location;
+use EzSystems\HybridPlatformUi\Decorator\LocationDecorator;
+
+class DecoratedLocationService
```
bdunogier: PhpDoc.

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    private $pathService;
+
+    public function __construct(LocationService $locationService, PathService $pathService)
+    {
+        $this->locationService = $locationService;
+        $this->pathService = $pathService;
+    }
+
+    public function loadLocations(ContentInfo $contentInfo)
```
bdunogier: PhpDoc.

src/lib/Decorator/ValueObjectDecorator.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Decorator;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
+abstract class ValueObjectDecorator
```
bdunogier: PhpDoc.

src/lib/Decorator/LocationDecorator.php
```diff
@@ -0,0 +1,59 @@
...
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Decorator;
+
+use eZ\Publish\API\Repository\Values\Content\Location;
+
+class LocationDecorator extends ValueObjectDecorator
```
bdunogier: PhpDoc.

bdunogier (changes requested):
Besides a couple comments, we're almost there.

As discussed on Slack, the main location should be shown first. With this added, we'll be able to merge this.

src/lib/Repository/PathService.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return Location[]
+     */
+    public function loadPathLocations(Location $location)
+    {
+        $locationQuery = new LocationQuery([
+            'filter' => new Ancestor($location->pathString),
+        ]);
+
+        return array_map(function (SearchHit $searchHit) {
```
bdunogier: Not sure about the CS / readability on this one :)

```
return array_map(
  function (SearchHit $searchHit) {
    return $searchHit->valueObject;
  },
  $this->searchService->findLocations($locationQuery)->searchHits
);
```

(not sure why you had extra parenthesis around the `findLocations()` call, though.

bdunogier (commented):
Forgot to post this one, sorry @mcurran-inviqa.

src/lib/Repository/PathService.php
bdunogier: P.S. as long as the cs script agrees, I'm not strongly set on that.

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,63 @@
...
+    private $pathService;
+
+    public function __construct(LocationService $locationService, PathService $pathService)
+    {
+        $this->locationService = $locationService;
+        $this->pathService = $pathService;
+    }
+
+    /**
+     * Load decorated locations.
```
bdunogier: I'd highlight a bit more that it loads _all_ the locations of a given ContentInfo, and decorates them with extra info. I'd _maybe_ even add which info.

There is an overhead when using this service. I'd really prefer if no developers use it without a good reason :)

bdunogier (changes requested):

bdunogier (commented):
I think we are almost good. We will evolve the services that were added when we work on other tabs.

bdunogier (approved):
Alright, let's go for it.

src/bundle/Resources/views/fields/location/path.html.twig
```diff
@@ -0,0 +1,8 @@
+{% spaceless %}
+    {% for location in locations %}
+        <a href="{{ path('_ezpublishLocation', {'locationId': location.id}) }}">{{ ez_content_name(location.contentInfo) }}</a>
```
dpobel: you can directly write:
```twig
<a href="{{ path(location) }}">
```

See https://doc.ez.no/display/DEVELOPER/ez_urlalias

src/bundle/Resources/views/fields/location/path.html.twig
```diff
@@ -0,0 +1,8 @@
...
+        {% if not loop.last %}
+        /
```
dpobel: instead of generating the separator here, I think it should be defined in the CSS and the markup should be something like:

```html
<ul class="ez-breadcrumbs">
    <li class="ez-breadcrumbs-item"><a href="/whatever">Parent</a></li>
    <li class="ez-breadcrumbs-item"><a href="/whatever/whatever2">Element</a></li>
</ul>
```

normally the corresponding CSS is already there

dpobel (changes requested):
besides the inline comments, I'm not super fan of the whole decorator thing. That feels a bit complicated, I was more expecting to have a dedicated controller/action able to generate the breadcrumb for a given Location. Also, because the breadcrumb of the current Location is supposed to be also visible outside of this tab (near the Content name) so a controller would make that super easy.

Anyway that point is more an open question for @bdunogier and @andrerom.
The *Request changes* status is actually related to the 2 comments in path.html.twig

src/bundle/Resources/config/services.yml
```diff
@@ -68,3 +68,13 @@ services:
         public: false
 
     EzSystems\HybridPlatformUi\Filter\VersionFilter: ~
+
+    ezsystems.platformui.hybrid.repository.path_service:
```
bdunogier: I'd change the definition to the autowired one:

```yaml
EzSystems\HybridPlatformUi\Repository\PathService:
  arguments:
    - "@ezpublish.api.service.inner_search"
```

src/bundle/Resources/config/services.yml
```diff
@@ -68,3 +68,13 @@ services:
...
+        class: EzSystems\HybridPlatformUi\Repository\PathService
+        arguments:
+            - "@ezpublish.api.service.inner_search"
+
+    EzSystems\HybridPlatformUi\Repository\DecoratedLocationService:
+        arguments:
+            - "@ezpublish.api.service.inner_location"
+            - "@ezsystems.platformui.hybrid.repository.path_service"
```
bdunogier: You can remove this one, it will be autowired thanks to typehint.

src/bundle/Controller/ContentViewController.php
```diff
@@ -86,6 +87,22 @@ public function versionsTabAction(ContentView $view, VersionFilter $versionFilte
         return $view;
     }
 
+    public function locationsTabAction(ContentView $view, DecoratedLocationService $decoratedLocationService)
```
bdunogier, line 90: I'd name the argument `$locationService`.

But as said on slack, I see the point where the name of the variable describes what it is. I'd say we're both right, and I'm not gonna ask for it.

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,79 @@
...
+                );
+
+                $decoratedLocation->main = ($location->id === $contentInfo->mainLocationId);
+
+                return $decoratedLocation;
+            },
+            $locations
+        );
+
+        foreach ($decoratedLocations as $key => $decoratedLocation) {
```
bdunogier, line 70: Could you refactor this block to a private method, so that we know at a glance what it does ?

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -0,0 +1,79 @@
...
+     *
+     * @return LocationDecorator[]
+     */
+    public function loadLocations(ContentInfo $contentInfo)
+    {
+        $locationService = $this->locationService;
+        $pathService = $this->pathService;
+        $locations = $locationService->loadLocations($contentInfo);
+
+        $decoratedLocations = array_map(
```
bdunogier, line 51: Could you refactor this block to a private method, so that we know at a glance what it does ?

bdunogier (changes requested):
Just a couple minor things, and we're good to go I think.

src/lib/Decorator/LocationDecorator.php
```diff
@@ -0,0 +1,72 @@
...
+     * Get value object.
+     *
+     * @return Location
+     */
+    public function getValueObject()
+    {
+        return $this->location;
+    }
+
+    public function isDraft()
```
maxbaldanza, line 58: I'm wondering if these methods are needed? Could you not have a `__call` method on the `ValueObjectDecorator` like you have for the `__get` methods that passes these to the valueobject? This would avoid having to add to these decorators when a new method was added to the API?

src/bundle/Resources/views/fields/location/path.html.twig
```diff
@@ -0,0 +1,5 @@
+<ul class="ez-breadcrumbs">
+    {% for location in locations %}
+    <li class="ez-breadcrumbs-item"><a href="{{ path(location) }}">{{ ez_content_name(location.contentInfo) }}</a></li>
+    {% endfor %}
+</ul>
```
maxbaldanza, line 5: Can we add a new line to this for consistency?

src/lib/Decorator/LocationDecorator.php
mcurran-inviqa: introspection from twig may have an issue with that.

bdunogier: As far as I know, @maxbaldanza, the issue is twig's introspection of the class that doesn't pick the parent's magic methods. Or something in this area.

maxbaldanza: Ah ok, does twig not take in to account the magic methods? Makes sense now

maxbaldanza (approved):
Good work

src/bundle/Resources/config/services.yml
```diff
@@ -68,3 +68,11 @@ services:
         public: false
 
     EzSystems\HybridPlatformUi\Filter\VersionFilter: ~
+
+    EzSystems\HybridPlatformUi\Repository\PathService:
+        arguments:
+            - "@ezpublish.api.service.inner_search"
```
bdunogier, line 74: Why did you use `@ezpublish.api.service.inner_search` instead of `@ezpublish.api.service.search` ?

src/lib/Repository/DecoratedLocationService.php
```diff
@@ -66,14 +76,15 @@ function (Location $location) use ($locationService, $pathService, $contentInfo)
             },
             $locations
         );
+    }
 
+    private function unshiftMainLocation(array &$decoratedLocations)
```
bdunogier: I'd go for a more semantic name. "unshift" is not _what_ the method does, it's _how_ it does it. What about "prioritizeMainLocation()", or something in this range ?

For consistency, I'd avoid the reference argument, and would make it return the newly organized array.

bdunogier (changes requested):

src/lib/Decorator/LocationDecorator.php
bdunogier: Not the parent's it seems.

bdunogier (approved):

bdunogier:
Asked @dpobel for an update of his review, then we an merge it.

dpobel (approved):

PR #23, opened 2017-06-19
[EZP-27359]: Move routes loading to HybridPlatformUI

lolautruche:
> https://jira.ez.no/browse/EZP-27359
> https://github.com/ezsystems/ezpublish-kernel/pull/2025

Routes for FOSJsRouting were loaded from a fragment in PlatformUIBundle.
Moved to HybridPlatformUI layout in order to get the right SiteAccess context

### DIFF
--- a/src/bundle/Resources/views/layout.html.twig
+++ b/src/bundle/Resources/views/layout.html.twig
@@ -34,16 +34,7 @@
     <link rel="stylesheet" href="{{ asset('bundles/ezsystemshybridplatformui/css/modules/list-toolbar.css') }}">
     <link rel="stylesheet" href="{{ asset('bundles/ezsystemshybridplatformui/css/modules/content-field.css') }}">
 
-    <script>
-        {#
-            FIXME:
-            // https://jira.ez.no/browse/EZP-27359
-            // this is a workaround to FOSJsRoutingBundle not generating URL for the admin SA
-            // of course, this only works when eZ Platform has one SA group and if the admin SA is not
-            // configured with a different matching strategy than URI...
-        #}
-        Routing.setBaseUrl('/admin');
-    </script>
+    <script src="{{ path('fos_js_routing_js', { callback: 'fos.Router.setData' }) }}"></script>
 
 {% if app.debug %}
     <script>


lolautruche:
Review ping @bdunogier @dpobel @andrerom

bdunogier (approved):

andrerom (approved):

src/bundle/Resources/views/layout.html.twig
```diff
@@ -34,16 +34,7 @@
...
-        {#
-            FIXME:
-            // https://jira.ez.no/browse/EZP-27359
-            // this is a workaround to FOSJsRoutingBundle not generating URL for the admin SA
-            // of course, this only works when eZ Platform has one SA group and if the admin SA is not
-            // configured with a different matching strategy than URI...
-        #}
-        Routing.setBaseUrl('/admin');
-    </script>
+    <script src="{{ path('fos_js_routing_js', { callback: 'fos.Router.setData' }) }}"></script>
```
dpobel, line 37: If I get the fix correctly, the point here is to include the exposed routes with a siteaccess aware URI that way, `Routing.generate` will generate correct URIs, right ?

dpobel (approved):
Note: after thinking a bit about it, I'm not super comfortable with this change because this *spreads* FOSJsRoutingBundle related code on both PlatformUIBundle and hybrid-platform-ui packages while only PlatformUIBundle would need it. At the moment `Routing.generate` is used in the PlatformUIBundle for the subitems and in hybrid-platform-ui for the Browse button but I think we should avoid that by changing the UDW to use it as well so that the UDW provides the URLAlias for the selected location(s). FOSJSRoutingBundle would then be used only by PlatformUIBundle code and then we'll have to find a way to move that back in PlatformUIBundle.
But for now, that's OK :)

src/bundle/Resources/views/layout.html.twig
lolautruche: Yes

lolautruche:
@dpobel I'm fine with having this in PlatformUIBundle only but obviously not loaded from a web component like it's currently the case, because that way we lose the knowledge of current siteaccess, and thus fallback to the default one.

andrerom:
So any changes needed to this?

dpobel:
as said in the end my review comment for now, that's ok like this but we need to wait for https://github.com/ezsystems/ezpublish-kernel/pull/2025 to merged otherwise this will break the JS generated URI.

PR #24, opened 2017-06-19
Added PHP CS Fixer configuration

bdunogier:
Title says it all.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,31 @@
+<?php
+return PhpCsFixer\Config::create()
+    ->setRules([
+        '@Symfony' => true,
+        '@Symfony:risky' => true,
+        'concat_space' => ['spacing' => 'one'],
+        'array_syntax' => false,
+        'simplified_null_return' => false,
+        'phpdoc_align' => false,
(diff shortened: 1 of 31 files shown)


bdunogier:
@andrerom how do we enable ezrobot for this ?

.php_cs
```diff
@@ -0,0 +1,61 @@
+<?php
+
+if (class_exists('Symfony\CS\Config\Config')) {
+    // PHP-CS-Fixer 1.x syntax (deprecated)
```
andrerom: don't need this, we only support one version and that is 2.x. the if cases where just needed during transition from 1.x ot 2.x

andrerom (changes requested):

andrerom:
> @andrerom how do we enable ezrobot for this ?

afaik you create a new job on jenkins, @vidarl might be able to show you how :)

andrerom (approved):

andrerom:
+1 assuming it passes ;)

bdunogier:
Okay, php-cs-fixer passed, merging

PR #27, opened 2017-06-22
EZP-27445: Use action autowiring rather than constructor in controller

maxbaldanza:


src/bundle/Resources/config/views.yml
```diff
@@ -13,5 +13,5 @@ system:
             versions_tab:
                 default:
                     template: "EzSystemsHybridPlatformUiBundle:content/tabs:versions.html.twig"
-                    controller: "EzSystemsHybridPlatformUiBundle:ContentView:versionsTab"
+                    controller: "EzSystems\\HybridPlatformUiBundle\\Controller\\ContentViewController:versionsTabAction"
```
bdunogier: I'd rather use single quotes to avoid escaping the backslash.

bdunogier (changes requested):
I'd change the definition of the VersionFilter service while at it.

Change:
```yaml
    ezsystems.platform_ui.hybrid.filter.version_filter:
        class: EzSystems\HybridPlatformUi\Filter\VersionFilter
```
to
```yaml
    EzSystems\HybridPlatformUi\Filter\VersionFilter: ~
```

src/bundle/Resources/config/views.yml
maxbaldanza: Sure, was just trying to keep it consistent but happy to update 👍

bdunogier: Valid argument, but I still think that we should go for single quotes. We could make it consistent by going single quotes everywhere. But then I'd prefer a distinct commit, if we can keep the PRs simple.

bdunogier (approved):
Great.

mcurran-inviqa (approved):

PR #29, opened 2017-06-22
EZP-27445: Add actions for draft and archived versions

maxbaldanza:
- [x] Add delete/new/edit actions to draft and archived versions
- [x] Handle submission of deleting a draft version

https://jira.ez.no/browse/EZP-27566
screen shot 2017-06-30 at 09 59 46


bdunogier:
We definitely need to use a custom controller for this. It is _really_ not a content _view_ thing :)

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -1,10 +1,18 @@
 {% trans_default_domain "locationview" %}
 
 {% if draftVersions %}
+    {{ form_start(draftActionsForm, {"action": path("ez_hybrid_platform_ui_content_view_draft_actions"), 'attr': {'class': 'ez-js-standard-form'}}) }}
+
     <div class="ez-list-toolbar">
         <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.draft.under.edit'|trans()|desc('Draft under edit') }}</h2>
+        <div class="container-buttons">
```
dpobel: the class should be `ez-list-toolbar-tools`

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -1,10 +1,18 @@
 {% trans_default_domain "locationview" %}
 
 {% if draftVersions %}
+    {{ form_start(draftActionsForm, {"action": path("ez_hybrid_platform_ui_content_view_draft_actions"), 'attr': {'class': 'ez-js-standard-form'}}) }}
```
dpobel: why putting `ez-js-standard-form` class ?

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -16,9 +24,18 @@
 {% endif %}
 
 {% if archivedVersions %}
+    {{ form_start(archivedActionsForm, {'attr': {'class': 'ez-js-standard-form'}}) }}
+
     <div class="ez-list-toolbar">
         <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.archived.versions'|trans()|desc('Archived versions') }}</h2>
+        <div class="container-buttons">
+            {{ form_row(archivedActionsForm.new_draft, {'label': 'locationview.versions.new.draft.selected.version'|trans()|desc('New draft based on selected version'), 'attr': {'class': 'ez-button ez-button-secondary'}}) }}
```
dpobel: I guess (from your screenshot) `form_row` does not just generate the button, is there a way to only have the button without any extra markup ?

src/bundle/Resources/views/content/tabs/versions/table.html.twig
```diff
@@ -32,6 +41,15 @@
             <td>
                 {{ version.modificationDate|localizeddate('medium', 'short', app.request.locale) }}
             </td>
+            {% if isDraft is defined %}
+                <td>
+                    <a href="#" class="table-iconbutton">
```
dpobel: that won't work (I know this comes from the UI guidelines...) If you really want to fill that column, for now just write something like:

```html
<td>
    <a href="#" class="ez-button" data-disabled>Edit</a>
</td>
```

dpobel (changes requested):

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: There looks to be some JS that listens to form submit and makes this an ajax request. To have it post in the normal way I've added this class: https://github.com/ezsystems/hybrid-platform-ui-core-components/blob/master/js/ez-platform-ui-app.js#L53

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: It wraps the submit in a div: `<div><button type="submit" id="archived_actions_new_draft" name="archived_actions[new_draft]" class="ez-button ez-button-secondary">New draft based on selected version</button></div>`
Does this outer div need removing?

src/bundle/Resources/views/content/tabs/versions/table.html.twig
maxbaldanza: I assumed some CSS would be added at a later point for these to be displayed and it would be fine to leave like this for now but I'll update to your suggested value

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: Symfony forms are quite extensible so it can be removed if needed: https://symfony.com/doc/current/form/form_customization.html
I presume it will need removing from all buttons throughout the repo?

src/bundle/Resources/views/content/tabs/versions.html.twig
dpobel: yeah that's the whole point of the app :) (btw that's documented in https://github.com/ezsystems/hybrid-platform-ui/blob/master/docs/howto/prevent_navigation_enhancement.md)
As you wrote, if you don't put it, the form is submitted in AJAX which should not be an issue but maybe you experienced some with that ?

src/bundle/Resources/views/content/tabs/versions.html.twig
dpobel: yes please, maybe `{{ form_widget(archivedActionsForm.new_draft) }}` would do the trick ?

bdunogier:
As said during the daily, the toggling of checkboxes is postponed to a later story.

The same feature is required by multiple tabs, and can be applied to all of them at once when we have it.

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -1,10 +1,18 @@
 {% trans_default_domain "locationview" %}
 
 {% if draftVersions %}
+    {{ form_start(draftActionsForm, {"action": path("ez_hybrid_platform_ui_content_view_draft_actions"), 'attr': {'class': 'ez-js-standard-form'}}) }}
+
     <div class="ez-list-toolbar">
         <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.draft.under.edit'|trans()|desc('Draft under edit') }}</h2>
+        <div class="ez-list-toolbar-tools">
+            {{ form_row(draftActionsForm.delete, {'label': 'locationview.versions.delete.selected.draft'|trans()|desc('Delete selected draft'), 'attr': {'class': 'ez-button ez-button-negative'}}) }}
```
SylvainGuittard: It should be `desc('Delete')` instead of `desc('Delete selected draft')` (_cf._ mockup)

src/bundle/Resources/views/content/tabs/versions.html.twig
```diff
@@ -16,9 +24,18 @@
 {% endif %}
 
 {% if archivedVersions %}
+    {{ form_start(archivedActionsForm, {'attr': {'class': 'ez-js-standard-form'}}) }}
+
     <div class="ez-list-toolbar">
         <h2 class="ez-list-toolbar-label">{{ 'locationview.versions.archived.versions'|trans()|desc('Archived versions') }}</h2>
+        <div class="ez-list-toolbar-tools">
+            {{ form_row(archivedActionsForm.new_draft, {'label': 'locationview.versions.new.draft.selected.version'|trans()|desc('New draft based on selected version'), 'attr': {'class': 'ez-button ez-button-secondary'}}) }}
+            {{ form_row(archivedActionsForm.delete, {'label': 'locationview.versions.remove.selected.archived.version'|trans()|desc('Remove selected archived version(s)'), 'attr': {'class': 'ez-button ez-button-negative'}}) }}
```
SylvainGuittard: It should be `desc('Delete')` instead of `desc('Remove selected archived version(s)')` (_cf._ mockup)

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: No `form_widget` has the same output. As far as I'm aware we'd have to customise that template. I'll put something together 😄

src/bundle/Resources/views/content/tabs/versions.html.twig
maxbaldanza: Is this the right class? I've added this class and the buttons show on top of each other as per my screenshot but when I remove the class they display side by side?

maxbaldanza: Ignore my previous comment 😄 I think the issue was my browser window was too small and therefore it was stacking them on top of each other.

src/bundle/Form/Versions/Data.php
```diff
@@ -0,0 +1,24 @@
...
+use Symfony\Component\Form\Extension\Core\Type\CheckboxType;
+use Symfony\Component\Form\Extension\Core\Type\CollectionType;
+use Symfony\Component\Form\FormBuilderInterface;
+
+abstract class Data extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('versionIds', CollectionType::class, array(
```
bdunogier: Short array syntax.

src/bundle/Controller/ContentViewController.php
```diff
@@ -72,8 +74,12 @@ public function detailsTabAction(ContentView $view)
         return $view;
     }
 
-    public function versionsTabAction(ContentView $view, VersionFilter $versionFilter)
-    {
+    public function versionsTabAction(
```
bdunogier, line 76: I think it would make sense to move the controller action to the `VersionController`.

src/lib/Form/UiFormFactory.php
```diff
@@ -0,0 +1,46 @@
...
+    {
+        $this->formFactory = $formFactory;
+    }
+
+    /**
+     * Create form to be used for draft actions on versions tab.
+     *
+     * @return \Symfony\Component\Form\FormInterface
+     */
+    public function createVersionsDraftActionForm()
```
bdunogier: Would it make sense to provide the initialization data to the factory methods ? In that case, the array of versions.

The benefit of the FormFactory is quite limited in the current fashion, is it ?

src/bundle/Controller/ContentViewController.php
```diff
@@ -88,12 +94,22 @@ public function versionsTabAction(ContentView $view, VersionFilter $versionFilte
             $translations[$version->id] = $this->getTranslations($version);
         }
 
+        $draftVersions = $versionFilter->filterDrafts($versions);
+        $draftActionsForm = $formFactory->createVersionsDraftActionForm();
+        $draftActionsForm->setData($versionMapper->mapToForm($draftVersions, $contentInfo));
```
bdunogier: `$contentInfo` isn't an argument of `mapToForm()`.

src/bundle/Controller/ContentViewController.php
```diff
@@ -88,12 +94,22 @@ public function versionsTabAction(ContentView $view, VersionFilter $versionFilte
...
+
+        $archivedVersions = $versionFilter->filterArchived($versions);
+        $archivedActionsForm = $formFactory->createVersionsArchivedActionForm();
+        $archivedActionsForm->setData($versionMapper->mapToForm($archivedVersions, $contentInfo));
```
bdunogier: `$contentInfo` isn't an argument of `mapToForm()`.

bdunogier (changes requested):

src/bundle/Controller/ContentViewController.php
maxbaldanza: I think it would be a good idea but this PR is already quite big and is holding up some of the other PR's. We would also need to refactor some of it because it uses this `$this->getTranslations` and `$this->loadUser`. Can it be done afterwards in a separate PR?

src/lib/Repository/VersionService.php
```diff
@@ -0,0 +1,40 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\ContentService;
+
+/**
+ * Service for allowing deletion of versions.
+ */
+class VersionService
```
andrerom: Like in #39 this should be named `UiVersionService` now right?

maxbaldanza: I'm not sure, I personally don't like the idea of prefixing classes unless they are needed. The reason the others are prefixed is to not confuse them with the ones in the kernel. As a `Version` service does not exist it doesn't technically need it. Do you think it's better to prefix them all for consistency or just the ones that are needed? 
I suppose if we prefix them all then it would avoid conflicts if one was added in the kernel at a later date?

mcurran-inviqa (approved):

src/lib/Repository/VersionService.php
andrerom: > I suppose if we prefix them all then it would avoid conflicts if one was added in the kernel at a later date?

ok with me

maxbaldanza: 👍 Already updated as I think that makes the most sense

src/bundle/Controller/ContentViewController.php
andrerom: It can be done afterwards, but please then do that tomorrow in sync with @bdunogier

andrerom (approved):

andrerom:
Merged to unblock the other PR's, @maxbaldanza please take care about feedback from @bdunogier tomorrow like the action move to VersionController.

PR #30, opened 2017-06-22
EZP-27543: Integrate PJAX notification in Hybrid Platform UI App

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27543
Requires: https://github.com/ezsystems/PlatformUIBundle/pull/881

# Description

This patch adds the parsing of the notifications generated in PJAX responses. As result, notifications after actions in the admin panel are back in the interface.

Screencast:

[](http://www.youtube.com/watch?v=NsfR1M-wacE "Click to play on Youtube.com")

# Tests

manual test + unit tests

### DIFF
--- a/src/lib/Pjax/PjaxResponseMainContentMapper.php
+++ b/src/lib/Pjax/PjaxResponseMainContentMapper.php
@@ -5,6 +5,7 @@ namespace EzSystems\HybridPlatformUi\Pjax;
 use EzSystems\HybridPlatformUi\Components\App;
 use EzSystems\HybridPlatformUi\Components\MainContent;
 use EzSystems\HybridPlatformUi\Mapper\MainContentMapper;
+use EzSystems\PlatformUIBundle\Notification\Notification;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
@@ -16,6 +17,17 @@ class PjaxResponseMainContentMapper implements MainContentMapper
      * @var \EzSystems\HybridPlatformUi\Components\App
      */
     private $app;
+
+    /**
+     * A map of PlatformUIBundle's notification state to Hybrid Platform UI
+     * notification type.
+     *
+     * @const Array
+     */
+    const STATE_TO_TYPE_MAP = [
+        Notification::STATE_STARTED => 'processing',
+        Notification::STATE_DONE => 'positive',
+    ];
 
     public function __construct(App $app)
     {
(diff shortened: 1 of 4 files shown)


src/lib/Pjax/PjaxResponseMainContentMapper.php
```diff
@@ -50,14 +62,46 @@ public function map($response)
...
+     * Notification HTML elements.
+     *
+     * @param \DOMNodeList $elements
+     * @return array
+     */
+    private function buildNotifications(\DOMNodeList $elements)
+    {
+        $notifications = [];
+
+        foreach ($elements as $element) {
```
bdunogier, line 86: Overall, a `notificationMapper` would make the spec much easier to write. I'm just not sure it's worth doing it now. What do you think @dpobel ?

src/lib/Pjax/PjaxResponseMainContentMapper.php
```diff
@@ -50,14 +62,46 @@ public function map($response)
...
+            $state = $element->getAttribute('data-state');
+            $type = isset(static::STATE_TO_TYPE_MAP[$state]) ? static::STATE_TO_TYPE_MAP[$state] : $state;
+            $timeout = 10;
+
+            if ($type === 'error') {
+                $timeout = 0;
+            }
+
+            $notifications[] = [
```
bdunogier, line 95: Would a Notification class make sense here ?

bdunogier (approved):
Overall good. Made suggestions to improve it a bit, and simplify testing, but I don't want to make them hard requirements.

src/lib/Pjax/PjaxResponseMainContentMapper.php
dpobel: clearly yes but if you don't mind, I'll let you (or someone else) do that in https://jira.ez.no/browse/EZP-27476 (SF error handling) where we'll have the same need :)

src/lib/Pjax/PjaxResponseMainContentMapper.php
dpobel: it could be done this way but on the other hand, the notifications are really a part of the PJAX response so it also makes sense to do that in the PjaxResponseMainContentMapper. But the parsing becomes more complicated and this class grows (that's unlikely), then yeah it would be a good refactoring.

PR #31, opened 2017-06-22
EZP-27532 translations tab

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27532
image

### DIFF
--- a/src/bundle/Resources/views/locationview.html.twig
+++ b/src/bundle/Resources/views/locationview.html.twig
@@ -14,6 +14,7 @@
             <li class="ez-tabs-label"><a href="#ez-tab-versions">{{ 'tab.versions'|trans|desc('Versions') }}</a></li>
             <li class="ez-tabs-label"><a href="#ez-tab-locations">{{ 'tab.locations'|trans|desc('Locations') }}</a></li>
             <li class="ez-tabs-label"><a href="#ez-tab-relations">{{ 'tab.related.content'|trans|desc('Related content') }}</a></li>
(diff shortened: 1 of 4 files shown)


mcurran-inviqa:
https://jira.ez.no/browse/EZP-27532
image

maxbaldanza (approved):
👍 for going back and adding `desc()` to the tabs

maxbaldanza:
Probably not for this ticket but putting this out there for suggestions: what are your thoughts on splitting up these actions in to separate controllers so they all handle one single responsibility?
That way the controller doesn't become over 200 lines of code and we can follow something close to the ADR (action domain responder) pattern

src/bundle/Resources/views/locationview.html.twig
```diff
@@ -14,6 +14,7 @@
             <li class="ez-tabs-label"><a href="#ez-tab-versions">{{ 'tab.versions'|trans|desc('Versions') }}</a></li>
             <li class="ez-tabs-label"><a href="#ez-tab-locations">{{ 'tab.locations'|trans|desc('Locations') }}</a></li>
             <li class="ez-tabs-label"><a href="#ez-tab-relations">{{ 'tab.related.content'|trans|desc('Related content') }}</a></li>
+            <li class="ez-tabs-label"><a href="#ez-tab-translations">{{ 'tab.translations'|trans()|desc('Translations') }}</a></li>
```
bdunogier: You can skip the `()` in the `trans` call.

bdunogier (approved):
Good to go.

PR #34, opened 2017-06-23
EZP-27437: Remove Browse component and update browse.php to return correct locationId

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27437

## Description

Removing browse component in favor of ez-browse custom element
Also updated browse.php to return correct null locationId if no starting location id defined


dpobel (approved):

src/bundle/Resources/views/layout.html.twig
```diff
@@ -13,6 +13,7 @@
 
     <link rel="import" href="{{ asset( 'assets/ezplatform/hybrid-platform-ui-core-components/core-components.html' ) }}">
     <link rel="import" href="{{ asset( 'bundles/ezsystemshybridplatformui/webcomponents/ez-browse.html' ) }}">
+    <link rel="import" href="{{ asset( 'bundles/ezsystemshybridplatformui/webcomponents/ez-server-side-content.html' ) }}">
```
dpobel: please do not add that, it was removed on purpose

src/lib/Components/Browse.php
```diff
@@ -21,8 +21,6 @@ protected function getLocationId()
         if ($this->request->attributes->get('location')) {
             return $this->request->attributes->get('location')->id;
         }
-
-        return 'false';
```
dpobel: that's a bit short, this method should always return *something* and the `_toString` and `jsonSerialize` should be updated accordingly

dpobel (changes requested):
also please update the description and the title of this PR as its purpose has changed

dpobel (approved):

PR #36, opened 2017-06-23
use inheritance over decoration for locations

mcurran-inviqa:
refactored locations decorator to use inheritance instead


src/lib/Repository/Values/Content/UiLocation.php
```diff
@@ -0,0 +1,42 @@
...
+    protected $pathLocations;
+
+    /**
+     * Main flag.
+     *
+     * @var bool
+     */
+    protected $main;
+
+    public function __construct(array $properties = [], Location $location)
```
bdunogier: `$location` needs to be made first, as it is not optional. It also feels more correct :)

src/lib/Repository/UiLocationService.php
```diff
@@ -0,0 +1,87 @@
...
+
+        $uiLocations = $this->makeUiLocations($locations, $contentInfo);
+        $uiLocations = $this->prioritizeMainLocation($uiLocations);
+
+        return $uiLocations;
+    }
+
+    private function makeUiLocations(array $locations, ContentInfo $contentInfo)
+    {
+        $locationService = $this->locationService;
```
bdunogier: I wouldn't create local var for properties. `$this` doesn't really complexify the code.

bdunogier (changes requested):
I think I like this version better.

Could you try to convince git that `DecoratedLocation` was _renamed_ to `UiLocation` (instead of delete / create) ?

src/lib/Repository/UiLocationService.php
mcurran-inviqa: it's because it's passed into another function and is no longer in scope.

bdunogier (approved):
Works for me. @maxbaldanza ?

maxbaldanza (approved):
LGTM 👍

bdunogier:
Pushed a couple changes to the PR.

As for typehinting, we need to typehint on the API classes & interfaces, _not Core_.

And as for `$this` inside anonymous functions, it was added in PHP 5.4.

src/lib/Repository/Values/Content/UiLocation.php
```diff
@@ -0,0 +1,43 @@
...
+     * Child count.
+     *
+     * @var int
+     */
+    protected $childCount;
+
+    /**
+     * Path locations.
+     *
+     * @var \eZ\Publish\API\Repository\Values\Content\Location[]
```
maxbaldanza, line 28: Could you not use `APILocation` here instead as it's already imported?

maxbaldanza (changes requested):

src/lib/Repository/Values/Content/UiLocation.php
bdunogier: We tend to use the FQN in PhpDoc. When you're not using a smart IDE, for instance when reviewing code, it can be really helpful.

src/lib/Repository/UiLocationService.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $locations = $this->locationService->loadLocations($contentInfo);
+
+        $uiLocations = $this->makeUiLocations($locations, $contentInfo);
+        $uiLocations = $this->prioritizeMainLocation($uiLocations);
+
+        return $uiLocations;
+    }
+
+    private function makeUiLocations(array $locations, ContentInfo $contentInfo)
```
bdunogier: I'd prefer `buildUiLocations`, for consistency with the kernel.

maxbaldanza (approved):

PR #37, opened 2017-06-23
EZP-27443: Display main content tab

maxbaldanza:
https://jira.ez.no/browse/EZP-27443

screen shot 2017-06-23 at 15 18 00

### DIFF
--- a/src/bundle/Resources/views/content/tabs/content.html.twig
+++ b/src/bundle/Resources/views/content/tabs/content.html.twig
@@ -1,0 +1,18 @@
+{% for group in fieldGroups %}
+    <section class="ez-fieldgroup">
+        <h3 class="ez-fieldgroup-name">{{ group.name }}</h3>
+
+        {% for fieldDefinition in group.fieldDefinitions %}
+            <div class="ez-content-field">
+                <p class="ez-content-field-name">{{ ez_field_name(content, fieldDefinition.identifier) }}:</p>
+                <div class="ez-content-field-value">
+                    {% if ez_is_field_empty(content, fieldDefinition.identifier) %}
+                        <em>{{"fieldview.field.empty"|trans({}, "fieldview")|desc('This field is empty')}}</em>
+                    {% else %}
+                        {{ ez_render_field(content, fieldDefinition.identifier) }}
+                    {% endif %}
+                </div>
+            </div>
+        {% endfor %}
+    </section>
+{% endfor %}
--- a/src/lib/Repository/UiFieldGroupService.php
+++ b/src/lib/Repository/UiFieldGroupService.php
@@ -1,0 +1,64 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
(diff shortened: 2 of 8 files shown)


src/bundle/Resources/views/content/tabs/content.html.twig
```diff
@@ -0,0 +1,22 @@
+{% set languageCode = content.versionInfo.contentInfo.mainLanguageCode %}
+
+{% for contentTypeGroup in contentType.contentTypeGroups %}
```
bdunogier: Not the right element :-)

This is the group(s) the Content Type is part of. You want to iterate on the FieldDefinitions groups, that can be accessed with `$fieldDefinition->fieldGroup`.

You can use `@ezpublish.fields_groups.list` to get the human readable strings (uses translations).

bdunogier (changes requested):

bdunogier:
Unfortunately, large changes are required, since the wrong domain objects were used.

src/bundle/Resources/views/content/tabs/content.html.twig
maxbaldanza: 🤔 if you look at `UiContentType:: getFieldDefinitionsByFieldGroup` I am using the `$fieldDefinition->fieldGroup` I thought the translated string could be obtained from the content type but I shall take a look at the mentioned service 👍

src/bundle/Resources/views/content/tabs/content.html.twig
```diff
@@ -0,0 +1,22 @@
...
+    {% set contentTypeName = contentTypeGroup.name(languageCode) %}
+
+    <section class="ez-fieldgroups ez-fieldgroup ez-fieldgroup">
```
dpobel: too many classes from v1 here :)
For now, the following markup should be enough:
```html
<section class="ez-fieldgroup">
    <h3 class="ez-fieldgroup-name">{{ fieldgroupName }}</h3>
    <!-- fields here -->
</section>
```

src/lib/Repository/UiFieldGroupService.php
```diff
@@ -0,0 +1,64 @@
...
+     * @return UiFieldGroup[]
+     */
+    public function loadFieldGroups(ContentInfo $contentInfo)
+    {
+        $contentType = $this->contentTypeService->loadContentType($contentInfo->contentTypeId);
+        $fields = $contentType->getFieldDefinitions();
+
+        $fieldGroups = [];
+        foreach ($this->fieldsGroupsList->getGroups() as $groupId => $groupName) {
+            $fieldsInGroup = array_filter($fields, function (FieldDefinition $field) use ($groupId) {
```
bdunogier: Would you mind refactoring this part to a private method ? It has quite a lot of logic and not much explanations. Something like `$fieldsInGroup = filterFieldsFromGroup($fields, $groupId);`.

bdunogier (changes requested):

dpobel (approved):

bdunogier (approved):

PR #38, opened 2017-06-26
Added the missing |desc calls to tabs labels

bdunogier:
Also regenerated translations with those strings.

### DIFF
--- a/src/bundle/Resources/translations/fieldview.en.xlf
+++ b/src/bundle/Resources/translations/fieldview.en.xlf
@@ -1,16 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-06-23T10:41:58Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-06-26T08:39:22Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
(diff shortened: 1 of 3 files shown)


src/bundle/Resources/views/locationview.html.twig
```diff
@@ -21,7 +23,7 @@
                     <p class="ez-content-field-name">{{ ez_field_name(content, field.fieldDefIdentifier) }}</p>
                     <div class="ez-content-field-value">
                         {% if ez_is_field_empty(content, field.fieldDefIdentifier) %}
-                            <em>{{"fieldview.field.empty"|trans({}, "fieldview")}}</em>
+                            <em>{{"fieldview.field.empty"|trans({}, "fieldview")|desc('The field is empty')}}</em>
```
maxbaldanza: Shouldn't it be `This field is empty`? That's what seems to be used on the mockups

src/bundle/Resources/translations/fieldview.en.xlf
```diff
@@ -1,16 +1,16 @@
...
-  <file date="2017-06-23T10:41:58Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-06-26T08:39:22Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
       <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
     </header>
     <body>
       <trans-unit id="977cb55c24b1708d46ccbd753c2cea745976ae21" resname="fieldview.field.empty">
-        <source>fieldview.field.empty</source>
+        <source>The field is empty</source>
```
maxbaldanza: `This field is empty`?

maxbaldanza (changes requested):

src/bundle/Resources/views/locationview.html.twig
bdunogier: Oh, yeah, thank you.

PR #39, opened 2017-06-26
EZP-27447: Display related content tab

maxbaldanza:
https://jira.ez.no/browse/EZP-27447

screen shot 2017-06-26 at 14 23 16


spec/Repository/UiRelationServiceSpec.php
```diff
@@ -0,0 +1,82 @@
...
+            ]
+        );
+        $contentInfo = new ContentInfo(['contentTypeId' => $contentTypeId]);
+
+        $contentService->loadRelations($versionInfo)->willReturn([
+            $relation,
+        ])->shouldBeCalled();
+
+        $versionInfo->getContentInfo()->willReturn($contentInfo)->shouldBeCalled();
+        $contentTypeService->loadContentType($contentTypeId)->willReturn($contentType)->shouldBeCalled();
```
bdunogier: Unless I am mistaken, `willReturn()` involves `shouldBeCalled()`.

src/bundle/Resources/views/content/tabs/relations/table.html.twig
```diff
@@ -0,0 +1,40 @@
...
+                        {{ 'locationview.relations.type.content_level_relation'|trans()|desc('Content level relation') }}
+                    {% endif %}
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::EMBED') %}
+                        {{ 'locationview.relations.type.embed'|trans()|desc('Embed') }}
+                    {% endif %}
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::LINK') %}
+                        {{ 'locationview.relations.type.link'|trans()|desc('Link') }}
+                    {% endif %}
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::FIELD') %}
+                        {{ 'locationview.relations.type.attribute'|trans()|desc('Attribute') }}
```
SylvainGuittard: As mentioned in the jira issue (Label change):
> When the relation is field specific, then the value in the relation type column will read 'Field (%name_of_Field%)' and NOT 'Attribute (%name_of_Field%)'

So I think it should be 
`{{ 'locationview.relations.type.field'|trans()|desc('Field') }}`

spec/Repository/UiRelationServiceSpec.php
maxbaldanza: No PHPspec does not make this assumption. You need to explicitly say that they should be called

src/bundle/Resources/views/content/tabs/relations/table.html.twig
maxbaldanza: Thanks Sylvain 👍 you are correct and this was missed. I've updated the label as suggested 😄

spec/Repository/UiRelationServiceSpec.php
bdunogier: Okay, my bad :)
Well, actually: https://github.com/phpspec/prophecy/issues/201#issuecomment-125668658.

src/lib/Repository/Values/Content/UiRelation.php
```diff
@@ -0,0 +1,42 @@
...
+ * Extends original value object in order to provide additional fields.
+ */
+class UiRelation extends Relation
+{
+    /**
+     * Field definition.
+     *
+     * @var string
+     */
+    protected $fieldDefinition;
```
bdunogier: I'd change it to `fieldDefinitionName`. `FieldDefinition` alone has a strong meaning in our domain.

src/lib/Repository/Values/Content/UiRelation.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * The content type for the destination.
+     *
+     * @var string
+     */
+    protected $destinationContentType;
```
bdunogier: Same than above, `destinationContentTypeName`.

src/bundle/Controller/ContentViewController.php
```diff
@@ -103,6 +104,19 @@ public function locationsTabAction(ContentView $view, UiLocationService $uiLocat
...
     }
 
+    public function relationsTabAction(ContentView $view, UiRelationService $relationService)
+    {
+        $versionInfo = $view->getContent()->getVersionInfo();
+        $contentInfo = $versionInfo->getContentInfo();
+
+        $view->addParameters([
+            'relationList' => $relationService->loadRelations($versionInfo),
+            'reverseRelations' => $relationService->loadReverseRelations($contentInfo),
```
bdunogier: Consistency: either `relationsList`+ `reverseRelationsList`, or `relations` + `reverseRelations`.

src/lib/Repository/UiRelationService.php
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * Service for loading relations with additional data not provided by the original API.
+ */
+class UiRelationService
+{
+    /**
+     * @var ContentService
+     */
+    private $contentService;
+    /**
```
bdunogier: CS: line break after the previous property and the new one.

bdunogier (changes requested):
Just a couple CS remarks, besides that it looks good.

src/lib/Repository/UiRelationService.php
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    private function buildUiRelations(array $relations, ContentInfo $contentInfo)
+    {
+        return array_map(
+            function (Relation $relation) use ($contentInfo) {
+                $contentType = $this->loadContentType($contentInfo);
+                $destinationContentInfo = $relation->getDestinationContentInfo();
+
+                $fieldDefinition = $contentType->getFieldDefinition($relation->sourceFieldDefinitionIdentifier);
```
dpobel: there's also the case where the relation is not coming from a Field Definition but is made at *content level*
See https://github.com/ezsystems/ezpublish-kernel/blob/7.0/eZ/Publish/API/Repository/Values/Content/Relation.php#L21

src/bundle/Resources/views/content/tabs/relations/table.html.twig
```diff
@@ -0,0 +1,40 @@
...
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::EMBED') %}
+                        {{ 'locationview.relations.type.embed'|trans()|desc('Embed') }}
+                    {% endif %}
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::LINK') %}
+                        {{ 'locationview.relations.type.link'|trans()|desc('Link') }}
+                    {% endif %}
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::FIELD') %}
+                        {{ 'locationview.relations.type.field'|trans()|desc('Field') }}
+                    {% endif %}
+                    ({{ relation.fieldDefinitionName }})
```
dpobel: my previous comments about relation type means this is invalid for `COMMON` Relation type

src/bundle/Resources/views/content/tabs/relations/table.html.twig
```diff
@@ -0,0 +1,40 @@
...
+                <th>{{ 'locationview.relations.relation.type'|trans()|desc('Relation type') }}</th>
+            </tr>
+        </thead>
+        <tbody>
+        {% for relation in relationList %}
+            <tr>
+                <td><a href="{{ path(relation.destinationLocation) }}">{{ relation.destinationContentInfo.name }}</a></td>
+                <td>{{ relation.destinationContentTypeName }}</td>
+                <td>
+                    {% if relation.type == constant('eZ\\Publish\\API\\Repository\\Values\\Content\\Relation::COMMON') %}
```
dpobel: given you have a service that builds a UiRelation I think you could move that logic there. UiRelation could for instance have `relationName` property (feel free to suggest a better name) and then here you could write something like:

```twig
<td>
    {{ relation.relationName }}
    {% if relation.fieldDefinitionName %} {# or even relation.fieldRelation which would relation.isFieldRelation() #}
        ({{ relation.fieldDefinitionName }})
    {% endif %}
</td>
```

src/lib/Repository/UiRelationService.php
```diff
@@ -0,0 +1,102 @@
...
+        $reverseRelations = $this->contentService->loadReverseRelations($contentInfo);
+
+        return $this->buildUiRelations($reverseRelations, $contentInfo);
+    }
+
+    private function buildUiRelations(array $relations, ContentInfo $contentInfo)
+    {
+        return array_map(
+            function (Relation $relation) use ($contentInfo) {
+                $contentType = $this->loadContentType($contentInfo);
```
dpobel: shouldn't you load the Content Type of the destination or of the source of the relation depending on if you are building the relation list or the reverse relation list ?

src/lib/Repository/UiRelationService.php
```diff
@@ -0,0 +1,102 @@
...
+     * Loads reverse relations for a piece of content.
+     * Retrieves reverse relations and then sets the field definition and the content type name.
+     *
+     * @param ContentInfo $contentInfo
+     *
+     * @return UiRelation[]
+     */
+    public function loadReverseRelations(ContentInfo $contentInfo)
+    {
+        $reverseRelations = $this->contentService->loadReverseRelations($contentInfo);
```
dpobel: this one has a special policy, so the currently logged in user might not have the right to load the reverse relations and you need to handle that gracefully

dpobel (changes requested):

src/bundle/Resources/views/content/tabs/relations/table.html.twig
bdunogier: Is this comment on the right line, @dpobel ? I don't see how it related to the `constant` usage.

Or were you talking about a `type` property, that would contain the translation label ?

src/lib/Repository/UiRelationService.php
bdunogier: Very good catch @dpobel.

src/lib/Repository/Values/Content/UiRelation.php
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * The content type name for the destination.
+     *
+     * @var string
+     */
+    protected $destinationContentTypeName;
+
+    /**
+     * The location for the destination.
```
bdunogier: This should be made more specific. The location is probably the main one. I'm also not sure about the phrasing. "Main location of the destination content" ?

bdunogier (changes requested):

src/bundle/Resources/views/content/tabs/relations/table.html.twig
maxbaldanza: 🤔 @dpobel how would you then handle translations for this label? My thoughts is that services shouldn't handle output unless it comes specifically from a model. Hardcoding labels to be displayed on the front end doesn't seem like the right approach to me

bdunogier: What about this:

```
{% for relation in relationList %}
<tr>
    <td><a href="{{ path(relation.destinationLocation) }}">{{ relation.destinationContentInfo.name }}</a></td>
    <td>{{ relation.destinationContentTypeName }}</td>
    <td>{{ ez_ui_relation_type(relation) }}</td>
</tr>
{% endfor %}        
```

The Twig function does the type checking, and uses the translation service to return the translated version.

maxbaldanza: Happy to add a twig function to handle this, although it does slightly obscure what it is it's doing. The main question is how likely is this to be used elsewhere? If it's unlikely to be used elsewhere then is it pragmatic to add?

bdunogier: It's not more difficult to implement than twig code, and right now it obscures the twig template, I think ;)
I don't know if it will be reused, but a twig extension is I think a nice way to abstract this.

src/lib/Repository/Values/Content/UiRelation.php
bdunogier: Actually, the [story says](https://jira.ez.no/browse/EZP-27447?focusedCommentId=212682&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-212682) that we don't display the main location path anymore, but the content's name.

maxbaldanza: The content name is being used for the display. This is used to link to the piece of content: https://github.com/inviqa/hybrid-platform-ui/blob/4438948c96a709a0fa97e5c98adc1f56bde5725c/src/bundle/Resources/views/content/tabs/relations/table.html.twig#L15
Is there a better way to do this without the location?

src/lib/Repository/Values/Content/UiRelation.php
```diff
@@ -0,0 +1,53 @@
...
+     *
+     * @var Location
+     */
+    protected $relationLocation;
+
+    /**
+     * The name for the relation.
+     * This will either come from destinationContentInfo OR sourceContentInfo depending upon if reverse relation or normal relation.
+     *
+     * @var Location
```
bdunogier: `string`

bdunogier:
Ping @SylvainGuittard about the above.

src/bundle/Resources/views/content/tabs/relations/table.html.twig
```diff
@@ -0,0 +1,31 @@
...
+                        ({{ relation.relationFieldDefinitionName }})
+                    {% endif %}
+                </td>
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+{% else %}
+    <p class="ez-relations-box-list-no-content">
+        {{ 'locationview.relations.no.related.content'|trans()|desc('This content has no related content.') }}
```
bdunogier: Independently of the comment about splitting reverse relations out, this label is incorrect for reverse ones. It was something like "This object isn't used by any other" or something in that vein.

bdunogier (approved):

dpobel (approved):

PR #40, opened 2017-06-27
EZP-27448: Display url aliases tab

maxbaldanza:
- [X] Display system aliases
- [X] Display custom aliases

### Improvement 
Should the blank table be shown if none exist or a message be displayed? 

https://jira.ez.no/browse/EZP-27448

screen shot 2017-06-27 at 11 54 59

### DIFF
--- a/src/bundle/Controller/ContentViewController.php
+++ b/src/bundle/Controller/ContentViewController.php
@@ -103,6 +103,18 @@ class ContentViewController extends Controller
         return $view;
     }
 
+    public function urlsTabAction(ContentView $view)
+    {
+        $urlAliasService = $this->getRepository()->getURLAliasService();
+
+        $view->addParameters([
+            'system_urls' => $urlAliasService->listLocationAliases($view->getLocation(), false),
+            'url_aliases' => $urlAliasService->listLocationAliases($view->getLocation()),
+        ]);
+
+        return $view;
+    }
+
     public function translationsTabAction(ContentView $view)
     {
         $view->addParameters([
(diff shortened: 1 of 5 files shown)


bdunogier:
> Should the blank table be shown if none exist or a message be displayed?

A published content item _always_ has an UrlAlias. I'd hide the table completely if there are none, just to be protective, though.

maxbaldanza:
From my understanding they'd always have a system url but they wouldn't always have a url alias. Maybe the top table along with title should be hidden if that's the case?
Or maybe a message should be shown instead of the table to say `This content has no aliases`

src/bundle/Controller/ContentViewController.php
```diff
@@ -103,6 +103,18 @@ public function locationsTabAction(ContentView $view, UiLocationService $uiLocat
         return $view;
     }
 
+    public function urlsTabAction(ContentView $view)
+    {
+        $urlAliasService = $this->getRepository()->getURLAliasService();
+
+        $view->addParameters([
+            'system_urls' => $urlAliasService->listLocationAliases($view->getLocation(), false),
```
dpobel, line 111: those variables should be in camel case

dpobel (commented):
The story is a bit weird and the fact that it contains 3 lines of text does not help.

When using `listLocationAliases`, we can only distinguishes auto generated URL Aliases (ie URL Aliases generated for a Location) from URL Aliases that were manually added. Given that, I really don't known which are *system URLs* and which are *URL Aliases*.

So before doing any further work on that tab, our PM should first correctly define what is expected here.
(ping @SylvainGuittard and others that Github do not want to autocomplete ;))

SylvainGuittard:
It's definitely too early to work on this feature. PM has to define clearly what are the expectations on this tab. 
So I suggest to put this PR on hold before reviewing or going further.

bdunogier:
In that case @SylvainGuittard I'll close it. There's no point in keeping it open, and it can easily be reopened.

SylvainGuittard:
Sure.

PR #41, opened 2017-06-27
Improved styling in tabs and admin panel

dpobel:
Requires: https://github.com/ezsystems/PlatformUIBundle/pull/884

# Description

This patch adds a bit of CSS to improve styling in the newly added tabs and in the admin panel.

Screencast:

[](http://www.youtube.com/watch?v=3aMJivzVsVY "Click to play on Youtube.com")

### DIFF
--- a/src/bundle/Resources/public/css/modules/table-data.css
+++ b/src/bundle/Resources/public/css/modules/table-data.css
@@ -32,7 +32,7 @@
     display: inline-block;
 }
 
-.ez-table-data td:last-of-type {
+.ez-table-data-row-tool td:last-of-type {
     text-align: right;
 }
 
--- a/src/bundle/Resources/public/css/modules/content-field.css
+++ b/src/bundle/Resources/public/css/modules/content-field.css
@@ -1,5 +1,5 @@
 .ez-content-field {
-    margin: 1em;
+    margin: 1em 0;
 }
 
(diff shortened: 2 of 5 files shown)


src/bundle/Resources/public/css/modules/table-data.css
```diff
@@ -32,7 +32,7 @@
     display: inline-block;
 }
 
-.ez-table-data td:last-of-type {
+.ez-table-data-row-tool td:last-of-type {
```
SylvainGuittard, line 35: @inakijv added new styles for tables to the guidelines today. With those tables, I think we cover most of the tables in the administration interface.
https://ezsystems.github.io/ui-guidelines/#components-tables

`ez-table-data-row-tool` is more generic than `ez-table-checkboxes`. So +1
But we will have to find names and a simple way to manage table styles.

SylvainGuittard (approved):

src/bundle/Resources/public/css/modules/table-data.css
dpobel: > ez-table-data-row-tool is more generic than ez-table-checkboxes.

actually, that's 2 different things :) A table can have checkboxes and/or buttons on each row or none or only one of them. So if there's something special with table having checkboxes, there'll probably be `ez-table-data-row-checkbox` or something like that.

PR #43, opened 2017-06-28
[WIP] Services autowiring

bdunogier:
> Status: ready for review

Integrates auto-wiring to the package's services.

### Specialized interfaces

For classes that inherit from a common interface, such as `RequestMatcherInterface`, an intermediate interface is required to get autowiring to work in a natural way.

Example:
```
class PjaxRequestMatcher implements RequestMatcherInterface { }
```

With this structure, we can not autowire the construtor arguments with a `PjaxRequestMatcher` (we want to typehint on interfaces, not implementations). We therefore make `PjaxRequestMatcher` an interface, and add an implementation of it.

```
interface PjaxRequestMatcher extends RequestMatcherInterface { }

class HeaderPjaxRequestMatcher implements PjaxRequestMatcher { }
```

The services are defined so that the interface has a default implementation:
```
EzSystems\HybridPlatformUi\Pjax\PjaxRequestMatcher: '@EzSystems\HybridPlatformUi\Pjax\HeaderPjaxRequestMatcher'

EzSystems\HybridPlatformUi\Pjax\HeaderPjaxRequestMatcher: ~
```

### TODO
- [x] Autowire the components


src/lib/View/AppConfigCoreViewMainContentMapper.php
```diff
@@ -0,0 +1,40 @@
...
+     */
+    private $app;
+
+    public function __construct(App $app)
+    {
+        $this->app = $app;
+    }
+
+    /**
+     * @param \eZ\Publish\Core\MVC\Symfony\View\View $view
```
mcurran-inviqa, line 28: could probably use inheritDoc here

src/lib/View/CoreViewMainContentMapper.php
```diff
@@ -6,36 +6,8 @@
 namespace EzSystems\HybridPlatformUi\View;
 
 use eZ\Publish\Core\MVC\Symfony\View\View;
-use EzSystems\HybridPlatformUi\Components\App;
-use EzSystems\HybridPlatformUi\Components\MainContent;
 use EzSystems\HybridPlatformUi\Mapper\MainContentMapper;
 
-class CoreViewMainContentMapper implements MainContentMapper
+interface CoreViewMainContentMapper extends MainContentMapper
```
mcurran-inviqa: Docblock?

src/lib/Http/HybridRequestMatcher.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Http;
+
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
```
mcurran-inviqa: Docblock?

src/lib/Http/FormatAttributeHtmlFormatRequestMatcher.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Http;
+
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcher;
+
```
mcurran-inviqa, line 10: This class name is slightly confusing it sounds like a combination of a method and a class name

src/lib/Http/AdminRequestMatcher.php
```diff
@@ -0,0 +1,16 @@
...
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+
+interface AdminRequestMatcher extends RequestMatcherInterface
+{
+    /**
+     * @param array $routes
+     */
+    public function setExcludedRoutesPrefixes($routes);
+
```
mcurran-inviqa: this is defined in parent interface?

mcurran-inviqa (changes requested):

src/lib/Http/FormatAttributeHtmlFormatRequestMatcher.php
bdunogier: Well, it may sound like it, but "format" is a noun in that case (the "format" request attribute). Do you have an alternative in mind.

mcurran-inviqa: a docblock may suffice here to remove the confusion

bdunogier:
Taken care of your comments @mcurran-inviqa. Better ?

src/lib/View/AppConfigCoreViewMainContentMapper.php
bdunogier: The doc is supposed to be inherited from the parent without the need for `inheritDoc`.

maxbaldanza (commented):
There's a few new files been added here do they not need the copyright notice adding?

mcurran-inviqa (approved):

PR #44, opened 2017-06-28
EZP-27504: location view deletion form

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27504

image

### DIFF
--- a/spec/Repository/UiLocationServiceSpec.php
+++ b/spec/Repository/UiLocationServiceSpec.php
@@ -6,7 +6,6 @@
  */
 namespace spec\EzSystems\HybridPlatformUi\Repository;
 
-use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\Core\Base\Exceptions\ForbiddenException;
@@ -18,16 +17,20 @@ use Prophecy\Argument;
 
 class UiLocationServiceSpec extends ObjectBehavior
 {
-    function let(LocationService $locationService, ContentService $contentService, PathService $pathService)
+    function let(LocationService $locationService, PathService $pathService)
     {
-        $this->beConstructedWith($locationService, $contentService, $pathService);
+        $this->beConstructedWith($locationService, $pathService);
     }
 
     function it_loads_ui_locations_with_a_child_count(LocationService $locationService)
     {
         $contentInfo = new ContentInfo(['mainLocationId' => 1]);
 
-        $location = new Location(['id' => 2]);
+        $location = new Location([
+            'id' => 2,
+            'contentInfo' => $contentInfo,
+        ]);
+
         $locationService->loadLocations($contentInfo)->willReturn([$location]);
         $locationService->getLocationChildCount($location)->willReturn(1);
 
@@ -40,7 +43,11 @@ class UiLocationServiceSpec extends ObjectBehavior
     {
         $contentInfo = new ContentInfo(['mainLocationId' => 1]);
 
-        $location = new Location(['id' => 2]);
+        $location = new Location([
+            'id' => 2,
+            'contentInfo' => $contentInfo,
+        ]);
+
         $locationService->loadLocations($contentInfo)->willReturn([$location]);
         $locationService->getLocationChildCount($location)->willReturn(1);
         $pathService->loadPathLocations(Argument::type(Location::class))->willReturn([$location]);
@@ -50,65 +57,63 @@ class UiLocationServiceSpec extends ObjectBehavior
         $this->loadLocations($contentInfo)->shouldBeLike([$uiLocation]);
     }
 
-    function it_loads_ui_locations_with_a_main_flag(LocationService $locationService)
-    {
-        $contentInfo = new ContentInfo(['mainLocationId' => 1]);
-
-        $location = new Location(['id' => 1]);
-        $locationService->loadLocations($contentInfo)->willReturn([$location]);
-        $locationService->getLocationChildCount($location)->willReturn(1);
-
-        $uiLocation = new UiLocation($location, ['childCount' => 1, 'main' => true]);
-
-        $this->loadLocations($contentInfo)->shouldBeLike([$uiLocation]);
-    }
-
     function it_puts_the_main_location_first(LocationService $locationService)
     {
         $contentInfo = new ContentInfo(['mainLocationId' => 1]);
 
-        $location = new Location(['id' => 2]);
-        $mainLocation = new Location(['id' => 1]);
+        $location = new Location([
+            'id' => 2,
+            'contentInfo' => $contentInfo,
+        ]);
+
+        $mainLocation = new Location([
+            'id' => 1,
+            'contentInfo' => $contentInfo,
+        ]);
 
         $locationService->loadLocations($contentInfo)->willReturn([$location, $mainLocation, $location]);
         $locationService->getLocationChildCount($location)->willReturn(1);
         $locationService->getLocationChildCount($mainLocation)->willReturn(1);
 
-        $uiLocation = new UiLocation($location, ['childCount' => 1, 'main' => false]);
+        $uiLocation = new UiLocation($location, ['childCount' => 1]);
 
-        $mainLocationDecorator = new UiLocation($mainLocation, ['childCount' => 1, 'main' => true]);
+        $uiMainLocation = new UiLocation($mainLocation, ['childCount' => 1]);
 
-        $this->loadLocations($contentInfo)->shouldBeLike([$mainLocationDecorator, $uiLocation, $uiLocation]);
+        $this->loadLocations($contentInfo)->shouldBeLike([$uiMainLocation, $uiLocation, $uiLocation]);
     }
 
-    function it_deletes_locations(LocationService $locationService, ContentService $contentService)
+    function it_deletes_secondary_locations(LocationService $locationService)
     {
         $mainLocationId = 1;
-        $contentId = 1;
         $deleteLocationId = 2;
 
         $contentInfo = new ContentInfo(['mainLocationId' => $mainLocationId]);
-        $contentService->loadContentInfo($contentId)->willReturn($contentInfo);
 
-        $location = new Location(['id' => $deleteLocationId]);
+        $location = new Location([
+            'id' => $deleteLocationId,
+            'contentInfo' => $contentInfo,
+        ]);
+
         $locationService->loadLocation($deleteLocationId)->willReturn($location);
         $locationService->deleteLocation($location)->shouldBeCalled();
 
-        $this->deleteLocations([$deleteLocationId], $contentId);
+        $this->deleteSecondaryLocations([$deleteLocationId]);
     }
 
-    function it_does_not_delete_the_main_location(LocationService $locationService, ContentService $contentService)
+    function it_refuses_to_delete_the_main_location(LocationService $locationService)
     {
         $mainLocationId = 1;
-        $contentId = 1;
 
         $contentInfo = new ContentInfo(['mainLocationId' => $mainLocationId]);
-        $contentService->loadContentInfo($contentId)->willReturn($contentInfo);
 
-        $location = new Location(['id' => $mainLocationId]);
+        $location = new Location([
+            'id' => $mainLocationId,
+            'contentInfo' => $contentInfo,
+        ]);
+
         $locationService->loadLocation($mainLocationId)->willReturn($location);
         $locationService->deleteLocation($location)->shouldNotBeCalled();
 
-        $this->shouldThrow(ForbiddenException::class)->duringDeleteLocations([$mainLocationId], $contentId);
+        $this->shouldThrow(ForbiddenException::class)->duringDeleteSecondaryLocations([$mainLocationId]);
     }
 }
--- a/spec/Repository/UiLocationServiceSpec.php
+++ b/spec/Repository/UiLocationServiceSpec.php
@@ -6,8 +6,10 @@
  */
 namespace spec\EzSystems\HybridPlatformUi\Repository;
 
+use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\Core\Base\Exceptions\ForbiddenException;
 use eZ\Publish\Core\Repository\Values\Content\Location;
 use EzSystems\HybridPlatformUi\Repository\PathService;
 use EzSystems\HybridPlatformUi\Repository\Values\Content\UiLocation;
@@ -16,9 +18,9 @@ use Prophecy\Argument;
 
 class UiLocationServiceSpec extends ObjectBehavior
 {
-    function let(LocationService $locationService, PathService $pathService)
+    function let(LocationService $locationService, ContentService $contentService, PathService $pathService)
     {
-        $this->beConstructedWith($locationService, $pathService);
+        $this->beConstructedWith($locationService, $contentService, $pathService);
     }
 
     function it_loads_ui_locations_with_a_child_count(LocationService $locationService)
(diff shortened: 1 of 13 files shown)


src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -1,7 +1,11 @@
 {% trans_default_domain "locationview" %}
+{{ form_start(actionsForm, {'action': path('ez_hybrid_platform_ui_location_actions', {'contentId': content.id})}) }}
 
 <div class="ez-list-toolbar">
     <h2 class="ez-list-toolbar-label">{{ 'locationview.locations.content.locations'|trans()|desc('Content locations') }}</h2>
+    <div class="container-buttons">
+        {{ form_row(actionsForm.delete, {'label': 'locationview.locations.delete.locations'|trans()|desc('Delete selected locations'), 'attr': {'class': 'ez-button ez-button-negative'}}) }}
```
maxbaldanza: Should use `form_widget` for this rather than `form_row` so that it only outputs the button without the surrounding div

src/bundle/Resources/config/services.yml
```diff
@@ -49,3 +49,7 @@ services:
 
     EzSystems\HybridPlatformUi\View\CoreViewMainContentMapper:
         alias: 'EzSystems\HybridPlatformUi\View\AppConfigCoreViewMainContentMapper'
+
+    EzSystems\HybridPlatformUi\Mapper\Form\LocationMapper: ~
```
maxbaldanza: Is this needed with the auto wiring?

src/bundle/Controller/LocationController.php
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+
+        return new RedirectResponse(
+            $router->generate(
+                '_ez_content_view',
+                [
+                    'contentId' => $content->id,
+                    'layout' => 'true',
+                    'viewType' => 'full',
```
maxbaldanza: Isn't `full` the default so could be left out?

spec/Repository/UiLocationServiceSpec.php
```diff
@@ -78,4 +80,35 @@ function it_puts_the_main_location_first(LocationService $locationService)
...
+        $contentService->loadContentInfo($contentId)->willReturn($contentInfo);
+
+        $location = new Location(['id' => $deleteLocationId]);
+        $locationService->loadLocation($deleteLocationId)->willReturn($location);
+        $locationService->deleteLocation($location)->shouldBeCalled();
+
+        $this->deleteLocations([$deleteLocationId], $contentId);
+    }
+
+    function it_does_not_delete_the_main_location(LocationService $locationService, ContentService $contentService)
```
maxbaldanza: Doesn't `it_doesnt_delete_the_main_location` make more sense?

mcurran-inviqa: same thing, would rather have both words than missing punctuation.

maxbaldanza: Fair enough, I prefer using `doesnt` but that's just a preference 👍

maxbaldanza (approved):
Nice 💯

spec/Repository/UiLocationServiceSpec.php
bdunogier: I'd actually have used `it_refuses_to_delete_the_main_location`.

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -18,7 +22,7 @@
     {% if locations is defined %}
         {% for location in locations %}
         <tr>
-            <td></td>
+            <td>{{ form_widget(actionsForm.removeLocations[location.id], {'attr': {'disabled': location.main}}) }}</td>
```
dpobel: besides disabling the checkbox when it's the main Location, it should also be disabled if the current user does not have the right to delete the Location. I'm commenting in the template but I don't know where this should be handled

src/lib/Repository/UiLocationService.php
```diff
@@ -52,14 +63,38 @@ public function loadLocations(ContentInfo $contentInfo)
...
+        $contentInfo = $this->contentService->loadContentInfo($contentId);
+
+        foreach ($locationIds as $locationId) {
+            $location = $this->locationService->loadLocation($locationId);
+
+            if ($this->isMainLocation($location, $contentInfo)) {
+                throw new ForbiddenException('Main location cannot be deleted.');
+            }
+
+            $this->locationService->deleteLocation($location);
```
dpobel: in addition the comment about disabling checkboxes, this can throw an UnauthorizedException and we might have the case where this exception is thrown for one Location but others can be/have been deleted. So I guess, there should a try catch there.

src/lib/Repository/UiLocationService.php
```diff
@@ -52,14 +63,38 @@ public function loadLocations(ContentInfo $contentInfo)
...
+
     private function buildUiLocations(array $locations, ContentInfo $contentInfo)
     {
         return array_map(
             function (Location $location) use ($contentInfo) {
                 $properties = [
                     'childCount' => $this->locationService->getLocationChildCount($location),
                     'pathLocations' => $this->pathService->loadPathLocations($location),
-                    'main' => ($location->id === $contentInfo->mainLocationId),
+                    'main' => $this->isMainLocation($location, $contentInfo),
```
dpobel: `isMain` maybe ? and maybe UiLocation should have a `isMain` method instead of injecting that?

src/bundle/Controller/ContentViewController.php
```diff
@@ -8,10 +8,10 @@
  */
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
 use eZ\Publish\API\Repository\Exceptions\NotFoundException;
 use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\API\Repository\Values\Content\VersionInfo;
-use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
```
bdunogier: I'd leave that where it was. Alphabetically sorted :)

src/bundle/Controller/ContentViewController.php
```diff
@@ -111,16 +111,21 @@ public function versionsTabAction(
         return $view;
     }
 
-    public function locationsTabAction(ContentView $view, UiLocationService $uiLocationService)
-    {
+    public function locationsTabAction(
```
bdunogier: As for the versions pull-request, I'd move the action to the `LocationController`.

dpobel (changes requested):

src/bundle/Controller/LocationController.php
```diff
@@ -0,0 +1,48 @@
...
+            if ($actionsForm->get('delete')->isClicked()) {
+                $uiLocationService->deleteLocations($locationIds, $content->id);
+            }
+        }
+
+        return new RedirectResponse(
+            $router->generate(
+                '_ez_content_view',
+                [
+                    'contentId' => $content->id,
```
bdunogier, line 65: We should add the `locationId` argument: if we only specify the content id, it displays the main location, while the operation might have been triggered from any of the alternative ones.

The question is how to get it there.

src/lib/Repository/UiLocationService.php
```diff
@@ -52,14 +63,38 @@ public function loadLocations(ContentInfo $contentInfo)
...
+    /**
+     * Deletes locations.
+     * Throws exception if there is an attempt to delete the main location.
+     *
+     * @param array $locationIds
+     * @param int $contentId
+     *
+     * @throws ForbiddenException
+     */
+    public function deleteLocations(array $locationIds, int $contentId)
```
bdunogier: We may want to rename it to `deleteSecondaryLocations()`, as it won't delete a main location.

bdunogier (changes requested):
Not finished yet, but you can get started on these comments.

src/bundle/Controller/LocationController.php
mcurran-inviqa: the reason for not doing so was in case we are on an alternative location and delete it

src/lib/Repository/UiLocationService.php
bdunogier: +1 on `isMain`.

src/bundle/Controller/LocationController.php
bdunogier: Then I suggest we detect which one was deleted, and act accordingly (if the active one was deleted, redirect to the main).

Can also be a follow-up, but we need to create a story for it.

mcurran-inviqa: I have updated so if active deleted, redirect to main

bdunogier: I have explained that particular problem to @SylvainGuittard a couple minutes ago.

A problem we have here, that I have discussed with @dpobel, is that we ultimately want to refresh the location tab only. It means that the action wouldn't redirect to location view, but to the `location_tab` for the current location.

If a user is viewing an alternative location and deletes that alternative location using the location tab, we would need to redirect to a completely different locationview (the main one). @dpobel would rather not :)

A solution would be to block the currently viewed location from being deleted (it was done like that in Legacy I think).

bdunogier: Given our conversation with @dpobel, we will prevent deletion of the currently viewed location for the time being.

Handling a location change at this point has an impact on async tabs, and we have enough on our plate for the time being. Can you add it in the current PR @mcurran-inviqa ?

src/bundle/Controller/LocationController.php
```diff
@@ -0,0 +1,79 @@
...
+use EzSystems\HybridPlatformUi\Repository\UiLocationService;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RouterInterface;
+
+class LocationController extends Controller
+{
+    public function contentViewTabAction(
+        ContentView $view,
+        UiLocationService $uiLocationService,
```
bdunogier, line 24: To be consistent with what @maxbaldanza has done, this should probably be injected as a ctor argument (required throughout the whole controller).

But let's change that in a next story in this area.

src/bundle/Controller/LocationController.php
```diff
@@ -0,0 +1,79 @@
...
+            $view->addParameters([
+                'locations' => $locations,
+                'actionsForm' => $actionsForm->createView(),
+            ]);
+        }
+
+        return $view;
+    }
+
+    public function actionsAction(
```
bdunogier, line 43: That's a really bad name :)
No better one in mind though.

src/bundle/Resources/config/routing.yml
```diff
@@ -8,3 +8,8 @@ ez_hybrid_platform_ui_version_draft_actions:
     defaults:
         _controller: EzSystemsHybridPlatformUiBundle:Version:draftActions
     methods: [POST]
+ez_hybrid_platform_ui_location_actions:
+    path: /location/actions/{contentId}/{locationId}
```
bdunogier: Could you please change that to `/content/{contentId}/locations` ?
`POST` is enough to indicate the `action` part I think.

(I have asked for a similar change in one of the versions pull-requests)

bdunogier (changes requested):

src/bundle/Resources/config/routing.yml
bdunogier: The `locationId` parameter should be passed as a query param: `/location/actions/{contentId}?redirectLocationId={locationId}`

The action should redirect to that ID if given, or to the main if not given.

src/lib/Repository/UiLocationService.php
```diff
@@ -46,20 +57,43 @@ public function loadLocations(ContentInfo $contentInfo)
...
+     *
+     * @param array $locationIds
+     *
+     * @throws ForbiddenException
+     */
+    public function deleteSecondaryLocations(array $locationIds)
+    {
+        foreach ($locationIds as $locationId) {
+            $location = $this->locationService->loadLocation($locationId);
+            $uiLocation = new UiLocation($location);
```
bdunogier: I wouldn't create an incomplete `UiLocation` here. Writing `if ($location->id === $location->contentInfo->mainLocationId)` isn't more complicated.

If you'd rather avoid the code duplication, then you could:
- create a protected method `UiLocationService::isMain(Location $location)`, that does this comparison, and return a boolean
- use that method in `buildUiLocation()`
- use that method here

spec/Repository/UiLocationServiceSpec.php
```diff
@@ -48,34 +58,90 @@ function it_loads_ui_locations_with_a_path_location(LocationService $locationSer
...
+            'contentInfo' => $contentInfo,
+        ]);
 
-        $locationService->loadLocations($contentInfo)->willReturn([$location, $mainLocation, $location]);
+        $locationService->loadLocations($contentInfo)->willReturn([$location]);
         $locationService->getLocationChildCount($location)->willReturn(1);
-        $locationService->getLocationChildCount($mainLocation)->willReturn(1);
 
-        $uiLocation = new UiLocation($location, ['childCount' => 1, 'main' => false]);
+        $repository->canUser('content', 'manage_locations', $location->getContentInfo())->willReturn(true);
```
bdunogier: `canUser` is deprecated since 6.6. Use `PermissionResolver::canUser()` instead.

bdunogier (changes requested):

src/bundle/Controller/LocationController.php
bdunogier: One difficulty with that is that it changes the logic behind `UiLocation::canRemove()`. Each location needs to know about the currently viewed location, passed as a controller argument.

Checking if a location is the currently viewed one requires that we have this value for comparison, either in the controller, or the template, or in the service. In addition, it may be needed when removing locations, if the `UiLocationService::deleteLocations()` method is meant to enforce that logic.

### Idea A: Load the locations from the perspective of another Location
One suggestion, after digging through the code: change `loadLocations()` to load the locations _given a Location_ (the currently viewed one, that you get from `$view`. That way, we know which one the loading is done from, and we can use that to customize the location object. Then I'd also change `deleteSecondaryLocations()` to take the location we delete from as an argument, so that we can do the same check. This service is the one that manipulates locations from a UI's perspective, it is normal that it has the context it needs to implement the business logic.

### Idea B: distinguish the extra Location data from the Permissions part
We could also separate the `UiLocation` object into:
- `UiLocation`, with `pathLocations` and `childCount` properties
- `EditableLocation`, that wraps / extends a `UiLocation`, and adds the `canDelete` property and permissions things

SylvainGuittard: > A solution would be to block the currently viewed location from being deleted (it was done like that in Legacy I think)

I just tested on legacy. The user can delete the location currently viewed and the user is redirected to the parent.
The only location that you cannot delete on legacy is the main location.

dpobel: that's an interesting topic :) and indeed with the tab + async system we have to find a solution. Since we are at the very beginning of the tab + async implementation (and not about to release a stable version), I agree with @bdunogier, we can avoid that complicated question by not allowing to remove the currently viewed Location **for now**. Of course we need a new story for that problem (explaining the topic and pointing to that thread).
I have some ideas to overcome that issue but I need to think a bit about it

src/lib/Repository/UiLocationService.php
mcurran-inviqa: this is how it was setup originally @dpobel suggested a change

src/bundle/Controller/LocationController.php
bdunogier: @SylvainGuittard the API allows you to delete the mainLocation, based on the code I have read. It is supposed to set the first remaining location as the new main one. Not tested it myself though.

SylvainGuittard: > we need a new story for that problem (explaining the topic and pointing to that thread)

@dpobel : done! https://jira.ez.no/browse/EZP-27603

bdunogier:
It seems that https://github.com/ezsystems/hybrid-platform-ui/pull/44#discussion_r125497881 (route change) was forgotten. Follow-up: https://jira.ez.no/browse/EZP-27604.

bdunogier (approved):

dpobel (approved):

PR #46, opened 2017-06-30
EZP-27534 update location sort order and field

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27534
image

This PR moves the previously hard coded sort order form into a symfony form and introduces an action for applying the submitted form.

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -65,7 +65,7 @@ class LocationController extends TabController
         return $this->reloadTab('locations', $content->id, $redirectLocationId);
     }
 
-    public function updateLocationDefaultSortOrderAction(
+    public function updateDefaultSortOrderAction(
         Location $location,
         Content $content,
         Request $request,
@@ -76,12 +76,7 @@ class LocationController extends TabController
         $form->handleRequest($request);
 
         if ($form->isValid()) {
-            $updateStruct = $locationService->newLocationUpdateStruct();
-
-            $updateStruct->sortField = $form->get('sortField')->getData();
-            $updateStruct->sortOrder = $form->get('sortOrder')->getData();
-
-            $locationService->updateLocation($location, $updateStruct);
+            $locationService->updateLocation($location, $form->getData());
         }
 
         return $this->reloadTab('details', $content->id, $location->id);
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -8,7 +8,9 @@
  */
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\Content;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
 use EzSystems\HybridPlatformUi\Repository\UiLocationService;
@@ -61,5 +63,27 @@ class LocationController extends TabController
         }
 
         return $this->reloadTab('locations', $content->id, $redirectLocationId);
+    }
+
+    public function updateLocationDefaultSortOrderAction(
+        Location $location,
+        Content $content,
+        Request $request,
+        LocationService $locationService,
+        UiFormFactory $formFactory
+    ) {
+        $form = $formFactory->createLocationOrderingForm($location);
+        $form->handleRequest($request);
+
+        if ($form->isValid()) {
+            $updateStruct = $locationService->newLocationUpdateStruct();
+
+            $updateStruct->sortField = $form->get('sortField')->getData();
+            $updateStruct->sortOrder = $form->get('sortOrder')->getData();
+
+            $locationService->updateLocation($location, $updateStruct);
+        }
+
+        return $this->reloadTab('details', $content->id, $location->id);
     }
 }
--- a/src/bundle/Form/Locations/Ordering.php
+++ b/src/bundle/Form/Locations/Ordering.php
@@ -69,7 +69,7 @@ class Ordering extends AbstractType
      */
     protected function getSortFields($currentSortField)
     {
-        $defaultSortFields = [
+        $sortFields = [
             /** @Desc("Content name") */
             $this->translator->trans('sort.name', [], 'locationview') => Location::SORT_FIELD_NAME,
             /** @Desc("Priority") */
@@ -80,7 +80,7 @@ class Ordering extends AbstractType
             $this->translator->trans('sort.published', [], 'locationview') => Location::SORT_FIELD_PUBLISHED,
         ];
 
-        $otherSortFields = [
+        $unsupportedFields = [
             /** @Desc("Location path") */
             $this->translator->trans('sort.path', [], 'locationview') => Location::SORT_FIELD_PATH,
             /** @Desc("Content type identifier") */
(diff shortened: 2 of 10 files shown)


src/bundle/Controller/LocationController.php
```diff
@@ -68,4 +70,37 @@ public function actionsAction(
             )
         );
     }
+
+    public function updateSubItemOrderAction(
```
dpobel: I know the story title is misleading but you actually set the Location default sort order and sort field which is then used by _default_ by subitems so I think this should be called `updateLocationDefaultSortOrderField` or something like that

src/bundle/Controller/LocationController.php
```diff
@@ -68,4 +70,37 @@ public function actionsAction(
...
+        }
+
+        return new RedirectResponse(
+            $router->generate(
+                '_ez_content_view',
+                [
+                    'contentId' => $content->id,
+                    'locationId' => $location->id,
+                    'layout' => 'true',
+                    'viewType' => 'full',
```
dpobel: this should be `locations_tab`

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -79,19 +79,15 @@
...
+    <h2 class="ez-list-toolbar-label">{{ 'locationview.details.sub.items.sorting.order'|trans()|desc('Sub-items sorting order') }}</h2>
 </div>
 
-<p class="ez-subitem-ordering-details">{{ 'locationview.details.default.listing.of.subitems.by'|trans()|desc('Default listing of sub-items by') }}
-    <select class="ez-subitems-ordering-sort-type" disabled>
-        {% for fieldId, fieldLabel in ordering.sortFields %}
-        <option {% if location.sortField == fieldId %}selected{% endif %} value="{{ fieldId }}">{{ fieldLabel.key|trans()|desc(fieldLabel.default) }}</option>
-        {% endfor %}
-    </select>
+{{ form_start(orderingForm, {'action': path('ez_hybrid_platform_ui_location_update_sub_items_order', {'locationId': location.id, 'contentId': contentInfo.id})}) }}
```
dpobel: to be consistent with others forms, the `<form>` tag should start right before the `<div class="ez-list-toolbar">` element

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -79,19 +79,15 @@
...
+
+<p class="ez-subitem-ordering-details">{{ 'locationview.details.order.by'|trans()|desc('Order by') }}
+    {{ form_widget(orderingForm.sortField) }}
     in
```
dpobel: missing a small translatable string ;-)

dpobel (changes requested):

src/bundle/Controller/LocationController.php
```diff
@@ -62,4 +64,26 @@ public function actionsAction(
...
+        Content $content,
+        Request $request,
+        LocationService $locationService,
+        UiFormFactory $formFactory
+    ) {
+        $form = $formFactory->createLocationOrderingForm($location);
+        $form->handleRequest($request);
+
+        if ($form->isValid()) {
+            $updateStruct = $locationService->newLocationUpdateStruct();
```
bdunogier: This is a typical case where the form could be return a ValueObject (the update struct) instead of manipulating an array.

src/bundle/Controller/LocationController.php
```diff
@@ -62,4 +64,26 @@ public function actionsAction(
 
         return $this->reloadTab('locations', $content->id, $redirectLocationId);
     }
+
+    public function updateLocationDefaultSortOrderAction(
```
maxbaldanza: As this is in the `Location` controller do we need the keyword `Location` in the action name? I think `updateDefaultSortOrderAction` would be enough. @dpobel  what do you think?

src/bundle/Controller/LocationController.php
maxbaldanza: Could the `data_class` of the form be set to `LocationUpdateStruct` to avoid this?

maxbaldanza (changes requested):

bdunogier (changes requested):

src/lib/Form/UiFormFactory.php
```diff
@@ -82,4 +91,15 @@ public function createLocationsActionForm(array $locations = [])
...
         return $this->formFactory->create(LocationActions::class, $data);
     }
+
+    public function createLocationOrderingForm(Location $location)
+    {
+        $data = $this->orderingMapper->mapToForm($location);
+
+        return $this->formFactory->create(
+            Ordering::class,
+            $data,
```
maxbaldanza: As this is already being called on a multi line, just wondering if rather than setting `$data` you could call `mapToForm` directly here. Should still be pretty readable

maxbaldanza (approved):

src/bundle/Resources/views/content/tabs/details.html.twig
```diff
@@ -78,20 +78,16 @@
...
-    </select>
-    in
-    <select class="ez-subitems-sorting-order" disabled>
-        {% for orderId, orderLabel in ordering.sortOrders %}
-        <option {% if location.sortOrder == orderId %}selected{% endif %} value="{{ orderId }}">{{ orderLabel.key|trans()|desc(orderLabel.default) }}</option>
-        {% endfor %}
-    </select>
+<p class="ez-subitem-ordering-details">{{ 'locationview.details.order.by'|trans()|desc('Order by') }}
+    {{ form_widget(orderingForm.sortField) }}
+    {{ 'locationview.details.in'|trans()|desc('in') }}
```
bdunogier: Could you please prefix make `locationview.details.in` more understandable ? Without knowing the context, translators will have a hard time figuring it out.

A good start would be to include `order` as for the `by` one above. I'd maybe change `in` to direction in the key as well. I'd make that `sort_order`:
- `locationview.details.sort_order.by`
- `locationview.details.sort_order.direction`

src/bundle/Form/Locations/Ordering.php
```diff
@@ -0,0 +1,117 @@
...
+        ];
+    }
+
+    /**
+     * @param int|null $currentSortField
+     * @return array
+     */
+    protected function getSortFields($currentSortField)
+    {
+        $defaultSortFields = [
```
bdunogier: I'd just call that `sortField`. `default` in this context doesn't really make sense.

src/bundle/Form/Locations/Ordering.php
```diff
@@ -0,0 +1,117 @@
...
+            /** @Desc("Location depth") */
+            $this->translator->trans('sort.depth', [], 'locationview') => Location::SORT_FIELD_DEPTH,
+            /** @Desc("Content type name") */
+            $this->translator->trans('sort.content.type.name', [], 'locationview') => Location::SORT_FIELD_CLASS_NAME,
+        ];
+
+        if ($currentSortField) {
+            $otherKey = array_search($currentSortField, $otherSortFields);
+            if ($otherKey) {
+                $defaultSortFields[$otherKey] = $otherSortFields[$otherKey];
```
bdunogier: I'm surprised that you have figured out that this had to be done. Good job ! Where did you find the info btw ?

One thing, though: by reading the code, it is hard to figure out why this is done if you're not aware of the unsupported sort fields. I suggest that you we rename `$otherSortFields` to `$unsupportedFields`. It should be sufficient. Otherwise, you could go with `$this->addUnsupportedSortField($currentSortField, $sortFields)`, that adds the unsupported one if necessary.

Why do you use the key here, btw ? There is a risk that the key will be between `0` and `3`, and that an unsupported field will override one of the `$defaultSortFields` entries. I'd just use `$defaultSortFields[] = $otherSortFields[$otherKey]`.

bdunogier (changes requested):

src/bundle/Form/Locations/Ordering.php
mcurran-inviqa: reverse engineered the original js code,

form expects (label => value) format, all the others have implicilty defined labels so should not overwrite.

src/bundle/Form/Locations/Ordering.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        $resolver
+            ->setDefaults([
+                'current_sort_field' => null,
```
bdunogier, line 113: Would it make sense to provide the FormType with a `Location` object instead ?

bdunogier (approved):

PR #47, opened 2017-06-30
EZP-27565: Use ez-content-view custom element

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27565

## Description

Use the new ez-content-view custom element and remove the css used for the content-view class

### DIFF
--- a/src/bundle/Resources/public/css/modules/content-view.css
+++ b/src/bundle/Resources/public/css/modules/content-view.css
@@ -1,22 +1,22 @@
-.ez-content-view {
+ez-content-view {
     display: flex;
 }
 
-.ez-content-view .ez-content-view-column {
+ez-content-view .ez-content-view-column {
     flex: 1 0;
 }
 
-.ez-content-view .ez-action-bar {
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/css/modules/content-view.css
```diff
@@ -1,22 +0,0 @@
-.ez-content-view {
```
dpobel, line 1: this of course completely breaks the content view page. The point of those PR is to replace `<ez-server-side-content class="ez-content-view">` by `<ez-content-view>`, in this CSS that means:

```css
ez-content-view {
    display: fex;
}

ez-content-view .ez-content-view-column {
    flex: 1 0;
}
```

and so on

dpobel (changes requested):

src/bundle/Resources/public/css/modules/content-view.css
dpobel: ok just notice you did that in the other PR, the CSS should still be here.

dpobel (approved):

PR #49, opened 2017-06-30
Let the hybrid app renderer pass RedirectResponse through

bdunogier:
Prevents the Hybrid app rendering from rendering RedirectResponse, as they are handled by the frontend application.


bdunogier:
See #50

PR #51, opened 2017-07-03
EZP-27566: Refactor versions

maxbaldanza:
- [x] Refactor loading of translations and user from controller in to services.
- [x] Move content viewing of versions in to version controller.

### DIFF
--- a/src/bundle/Controller/VersionController.php
+++ b/src/bundle/Controller/VersionController.php
@@ -7,6 +7,8 @@
  */
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+use EzSystems\HybridPlatformUi\Filter\VersionFilter;
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
@@ -29,6 +31,31 @@ class VersionController extends TabController
     ) {
         $this->uiVersionService = $uiVersionService;
         parent::__construct($router, $contentService);
+    }
+
+    public function contentViewTabAction(
+        ContentView $view,
+        VersionFilter $versionFilter,
+        UiFormFactory $formFactory
+    ) {
+        $contentInfo = $view->getContent()->getVersionInfo()->getContentInfo();
+        $versions = $this->uiVersionService->loadVersions($contentInfo);
+
+        $draftVersions = $versionFilter->filterDrafts($versions);
+        $draftActionsForm = $formFactory->createVersionsDraftActionForm($draftVersions);
+
+        $archivedVersions = $versionFilter->filterArchived($versions);
+        $archivedActionsForm = $formFactory->createVersionsArchivedActionForm($archivedVersions);
+
+        $view->addParameters([
+            'draftVersions' => $draftVersions,
+            'publishedVersions' => $versionFilter->filterPublished($versions),
+            'archivedVersions' => $archivedVersions,
+            'draftActionsForm' => $draftActionsForm->createView(),
+            'archivedActionsForm' => $archivedActionsForm->createView(),
+        ]);
+
+        return $view;
     }
 
     public function draftActionsAction(
--- a/src/lib/Repository/UiTranslationService.php
+++ b/src/lib/Repository/UiTranslationService.php
@@ -1,0 +1,40 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\LanguageService;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+
+/**
+ * Service for loading translations.
+ */
+class UiTranslationService
+{
+    /**
+     * @var LanguageService
+     */
+    private $languageService;
+
+    public function __construct(LanguageService $languageService)
+    {
+        $this->languageService = $languageService;
+    }
+
+    /**
+     * Load translations for a version.
+     *
+     * @param VersionInfo $versionInfo
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Language[]
+     */
+    public function loadTranslations(VersionInfo $versionInfo)
+    {
+        return array_map(function ($languageCode) {
+            return $this->languageService->loadLanguage($languageCode);
+        }, $versionInfo->languageCodes);
+    }
+}
--- a/src/lib/Repository/UiUserService.php
+++ b/src/lib/Repository/UiUserService.php
@@ -1,0 +1,66 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\Exceptions\NotFoundException;
+use eZ\Publish\API\Repository\Repository;
+use eZ\Publish\API\Repository\UserService;
+
+/**
+ * Service for loading user.
+ */
+class UiUserService
+{
+    /**
+     * @var UserService
+     */
+    private $userService;
+
+    /**
+     * @var Repository
+     */
+    private $repository;
+
+    public function __construct(Repository $repository, UserService $userService)
+    {
+        $this->repository = $repository;
+        $this->userService = $userService;
+    }
+
+    /**
+     * Tries to load a user ignoring privileges.
+     * This is so that the user of a version/detail information can be displayed even if the user
+     * doesn't have the privileges to do so.
+     *
+     * @param mixed $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User|null
+     */
+    public function findUserById($userId)
+    {
+        return $this->repository->sudo(function () use ($userId) {
+            return $this->loadUser($userId);
+        });
+    }
+
+    /**
+     * Loads user and handles not found exception.
+     * Intended usage, is to enable testing using php spec as testing callable is difficult.
+     * Use findUserById().
+     *
+     * @param mixed $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User|null
+     */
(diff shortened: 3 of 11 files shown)


src/lib/Repository/UiSudoService.php
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\Repository;
+
+/**
+ * Service for loading translations.
```
bdunogier: Probably not :)

src/lib/Repository/UiSudoService.php
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    /**
+     * Loads a user ignoring privileges.
+     *
+     * @param int $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User
+     */
+    public function loadUser(int $userId)
```
bdunogier: We can't typehint this as an `int`. The API documents them as `mixed`, as they may be strings at some point.

src/lib/Repository/UiSudoService.php
```diff
@@ -0,0 +1,44 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\Repository;
+
+/**
+ * Service for loading translations.
+ */
+class UiSudoService
```
bdunogier: This won't end well. We may end up with many different methods in this class, for all services (since the name isn't specific to users).

Should we rather use a decorator for this ? Or make it part of the `UiUserService` ? After all, using `sudo()` is part of this layer's business logic.

bdunogier (changes requested):

src/lib/Repository/UiTranslationService.php
```diff
@@ -33,11 +33,8 @@ public function __construct(LanguageService $languageService)
      */
     public function loadTranslations(VersionInfo $versionInfo)
     {
-        $translations = [];
-        foreach ($versionInfo->languageCodes as $languageCode) {
-            $translations[] = $this->languageService->loadLanguage($languageCode);
-        }
-
-        return $translations;
+        return array_map(function ($languageCode) {
```
bdunogier: I personally find this syntax to be easier to read:
```
return array_map(
  function ($languageCode) {
    return $this->languageService->loadLanguage($languageCode);
  },
  $versionInfo->languageCodes
```

Not a request for change though.

src/bundle/Controller/VersionController.php
```diff
@@ -16,6 +18,32 @@
 
 class VersionController extends Controller
 {
+    public function contentViewTabAction(
```
bdunogier: _technically_ it's a location view tab (the view is routed from a location's URL alias).

src/lib/Repository/UiUserService.php
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    /**
+     * Loads a user ignoring privileges.
+     *
+     * @param mixed $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User
+     */
+    public function loadUserIgnoringPermissions($userId)
```
bdunogier: I have been bothered by `loadUser` returning `null`, but I couldn't figure out why.

In the Repository's API, `loadAnything()` _will_ throw a NotFound. It's on purpose. You're asking to _load_ something that can't be found. We ended up preferring `findStuff` when there might not be a result.

What about `findUser()` ? Or `findUserById()`.

Not a request for change for this PR though. Up to you, if you see a way to implement this that fits.

The `IgnoringPermissions` part leaves me a bit skeptical as well. The fact is that for what we're doing here, we always want to ignore permissions. We want those users names and links. Could we make that the default for this service ? We may have to name it a bit differently to reflect that, but if we consider that those services reflect the business logic of the frontend, it can make sense.

src/lib/Repository/UiUserService.php
```diff
@@ -0,0 +1,62 @@
...
+     * @param mixed $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User|null
+     */
+    public function loadUser($userId)
+    {
+        try {
+            return $this->userService->loadUser($userId);
+        } catch (NotFoundException $e) {
+            return null;
```
bdunogier, line 63: Damien suggested in another PR that system values such as `null` are defined using constants. I kind of like that as a good practice. What do you think ?

src/lib/Repository/UiVersionService.php
```diff
@@ -17,9 +20,38 @@ class UiVersionService
...
+    /**
+     * Load versions and adds the author and translations.
+     *
+     * @param ContentInfo $contentInfo
+     *
+     * @return UiVersionInfo[]
+     */
+    public function loadVersions(ContentInfo $contentInfo)
+    {
+        $versions = $this->contentService->loadVersions($contentInfo);
```
bdunogier: Style:

```
return $this->buildUiVersions(
  $versions = $this->contentService->loadVersions($contentInfo)
);
```

We avoid deep nesting, but that's really fine :)

src/lib/Repository/Values/Content/UiVersionInfo.php
```diff
@@ -0,0 +1,35 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Repository\Values\Content;
+
+use eZ\Publish\Core\Repository\Values\Content\VersionInfo;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo as APIVersionInfo;
+
+/**
+ * Extends original value object in order to provide additional fields.
+ * Takes a standard VersionInfo instance and retrieves properties from it in addition to the provided properties.
```
bdunogier, line 13: Damn, something I've forgotten since the beginning: we need to document the properties we add to the value objects:

```
/**
 * @property-read \eZ\Publish\API\Repository\Values\User\User $author
 */
```

Applies to all other objects of course :p

src/lib/Repository/UiTranslationService.php
```diff
@@ -0,0 +1,40 @@
...
+ */
+namespace EzSystems\HybridPlatformUi\Repository;
+
+use eZ\Publish\API\Repository\LanguageService;
+use eZ\Publish\API\Repository\Values\Content\VersionInfo;
+
+/**
+ * Service for loading translations.
+ */
+class UiTranslationService
```
bdunogier, line 15: I'm a bit skeptical at a separate service for this. Having language handling inside the `UiUserService` wouldn't make it _too_ complex, would it ?

bdunogier (changes requested):
A couple minor changes, but the main lines are good.

src/lib/Repository/UiTranslationService.php
maxbaldanza: What's the reason for being skeptical? I don't think it would make it too complex but it would mean that it no longer has a single responsibility as it would be for loading users and loading translations

src/lib/Repository/UiUserService.php
maxbaldanza: 🤔 I'm not sure what value is gained from that in this case. Doesn't this make it harder to read? A developer is used to methods potentially returning null but returning a constant that is used as null is not something that is widely used

src/lib/Repository/Values/Content/UiVersionInfo.php
maxbaldanza: Added 👍 I'll go through and create a separate PR that will add it to the other VO

maxbaldanza: Added to the others here: https://github.com/ezsystems/hybrid-platform-ui/pull/63

src/lib/Repository/UiTranslationService.php
mcurran-inviqa: this doesn't seem to be anything to do with users?

src/lib/Repository/UiUserService.php
mcurran-inviqa: null is a pretty standard response for a single item query that returns no result, having a constant seems to be making it more complex than it needs to be.

mcurran-inviqa (approved):

src/lib/Repository/UiTranslationService.php
bdunogier: I had seen it only used in the UserService.
Nevermind this :)

src/lib/Repository/UiUserService.php
bdunogier: These got mixed up a bit.

Our API, which this software is based upon, throws NotFound exceptions when a `load()` method is called, and doesn't find the requested item. We have our own standards. We call the ones who may not find what they're asked for `findSomething()`.

Since we have `findUserById()` above, and `loadUser()` is not used anymore, we can just remove it, and ignore the method altogether.

bdunogier (changes requested):
We can merge this once `loadUser()` has been removed.

src/lib/Repository/UiUserService.php
maxbaldanza: The reason for separating these methods is that php spec has difficulty testing callables. So `loadUser` is being used in the spec to make sure that we catch the `NotFoundException` that is thrown

bdunogier: Okay... Could you add that to the phpdoc then ?

src/lib/Repository/UiUserService.php
```diff
@@ -0,0 +1,64 @@
...
+        $this->repository = $repository;
+        $this->userService = $userService;
+    }
+
+    /**
+     * Loads a user ignoring privileges.
+     *
+     * @param mixed $userId
+     *
+     * @return \eZ\Publish\API\Repository\Values\User\User
```
bdunogier: phpdoc: This may return `null `. Also, the method's description should be updated to reflect the `find` approach (as opposed to `load`).

bdunogier (changes requested):
Minor phpdoc remark, besides that we're good.

bdunogier:
@maxbaldanza do you think you could rebase the PR to have two commits matching the tasks you have described in the PR ? I'd rather not just squash this all.

maxbaldanza:
@bdunogier rebased and ready for review

bdunogier (approved):

PR #52, opened 2017-07-04
EZP-27567: Allow deletion of archived versions

maxbaldanza:
https://jira.ez.no/browse/EZP-27567

### DIFF
--- a/src/bundle/Controller/VersionController.php
+++ b/src/bundle/Controller/VersionController.php
@@ -22,16 +22,21 @@ class VersionController extends Controller
      */
     private $uiVersionService;
 
-    public function __construct(UiVersionService $uiVersionService)
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    public function __construct(UiVersionService $uiVersionService, RouterInterface $router)
     {
         $this->uiVersionService = $uiVersionService;
+        $this->router = $router;
     }
 
     public function draftActionsAction(
         $contentId,
         Request $request,
-        UiFormFactory $formFactory,
-        RouterInterface $router
+        UiFormFactory $formFactory
     ) {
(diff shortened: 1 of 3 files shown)


bdunogier (approved):

bdunogier:
Not a huge fan of how this is factored, but it is consistent with what exists, and can be improved later.

src/bundle/Controller/VersionController.php
```diff
@@ -10,31 +10,68 @@
...
+        $selectedIds = array_keys($form->get('versionIds')->getData());
+
+        if ($form->get('delete')->isClicked()) {
+            foreach ($selectedIds as $versionId) {
+                $this->uiVersionService->deleteVersion((int) $contentId, $versionId);
             }
         }
+    }
+
+    protected function redirectUser(RouterInterface $router, $contentId)
```
bdunogier: I'd prefer `redirectToContentView()`.
Also, if the `Router` was injected into the class instead of the action, we could avoid passing it as an argument.

bdunogier (changes requested):

src/bundle/Resources/config/routing.yml
```diff
@@ -8,3 +8,8 @@ ez_hybrid_platform_ui_version_draft_actions:
     defaults:
         _controller: EzSystemsHybridPlatformUiBundle:Version:draftActions
     methods: [POST]
+ez_hybrid_platform_ui_version_archive_actions:
+    path: /version/archive-actions/{contentId}
```
bdunogier: I'd rather have `content` in the URI.
Maybe `/content/{contentId}/versions/archives` ?

bdunogier (approved):

bdunogier:
@mcurran-inviqa can you also review this please ?

mcurran-inviqa (approved):

PR #54, opened 2017-07-05
EZP-27505 swap location

maxbaldanza:
- [X] Allow swapping of two locations 

Until UDW is implemented than a input box has been used for the location id.

https://jira.ez.no/browse/EZP-27505

screen shot 2017-07-07 at 12 47 02

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -8,16 +8,16 @@
  */
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
-use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
+use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+use eZ\Publish\API\Repository\Values\Content\Location;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
 use EzSystems\HybridPlatformUi\Repository\UiLocationService;
-use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\RouterInterface;
 
-class LocationController extends Controller
+class LocationController extends TabController
 {
     /**
      * @var UiLocationService
@@ -29,10 +29,15 @@ class LocationController extends Controller
      */
     private $formFactory;
 
-    public function __construct(UiLocationService $uiLocationService, UiFormFactory $formFactory)
-    {
+    public function __construct(
+        UiLocationService $uiLocationService,
+        UiFormFactory $formFactory,
+        ContentService $contentService,
+        RouterInterface $router
+    ) {
         $this->uiLocationService = $uiLocationService;
         $this->formFactory = $formFactory;
+        parent::__construct($router, $contentService);
     }
 
     public function contentViewTabAction(
@@ -58,50 +63,43 @@ class LocationController extends Controller
 
     public function actionsAction(
         Content $content,
-        Request $request,
-        RouterInterface $router
+        Request $request
     ) {
         $actionsForm = $this->formFactory->createLocationsActionForm();
         $actionsForm->handleRequest($request);
+
+        $redirectLocationId = $request->query->get('redirectLocationId', $content->contentInfo->mainLocationId);
 
         if ($actionsForm->isValid()) {
             $locationIds = array_keys($actionsForm->get('removeLocations')->getData());
 
             if ($actionsForm->get('delete')->isClicked()) {
                 $this->uiLocationService->deleteLocations($locationIds);
+
+                if (in_array($redirectLocationId, $locationIds)) {
+                    return $this->resetLocation($content->id);
+                }
             }
         }
 
-        return new RedirectResponse(
-            $router->generate(
-                '_ez_content_view',
-                [
-                    'contentId' => $content->id,
-                    'locationId' => $request->query->get('redirectLocationId', null),
-                ]
-            )
-        );
+        return $this->reloadTab('locations', $content->id, $redirectLocationId);
     }
 
     public function swapLocationAction(
-        $contentId,
-        $locationId,
-        Request $request,
-        RouterInterface $router
+        Content $content,
+        Location $location,
+        Request $request
     ) {
         $swapLocationsForm = $this->formFactory->createLocationsContentSwapForm();
         $swapLocationsForm->handleRequest($request);
 
         if ($swapLocationsForm->isValid()) {
             $newLocationId = $swapLocationsForm->get('new_location_id')->getData();
-            $this->uiLocationService->swapLocations($locationId, $newLocationId);
+            $this->uiLocationService->swapLocations($location->id, $newLocationId);
         }
-        //@TODO Show success/fail message to user
-        return new RedirectResponse(
-            $router->generate(
-                '_ez_content_view',
-                ['contentId' => $contentId]
-            )
-        );
+
+        $redirectLocationId = $request->query->get('redirectLocationId', $content->contentInfo->mainLocationId);
+
+        return $this->reloadTab('locations', $content->id, $redirectLocationId);
     }
 }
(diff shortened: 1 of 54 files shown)


src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -39,3 +39,18 @@
...
+<div class="ez-locations-box">
+    <h2 class="ez-locations-box-title">{{ 'locationview.locations.swap.title'|trans()|desc('Location Content Swap') }}</h2>
+    {{ form_start(swapLocationsForm, {"action": path("ez_hybrid_platform_ui_location_swap_location", {'contentId': content.id, 'locationId': locationId})}) }}
+    <p>
+        {{ 'locationview.locations.swap.with.another'|trans()|desc('Swap the content item at this location with another') }}
+
+        {{ form_widget(swapLocationsForm.swap, {
+            'label': 'locationview.locations.select.content.item'|trans()|desc('Select content item'),
+            'attr': {'class': 'ez-button ez-button-primary'},
+            'icon': 'ez-icon-relations',
```
dpobel: the whole icon topic in Hybrid Platform UI has to be completely designed (https://jira.ez.no/browse/EZP-27407) so for now, it's better to just ignore that until we have decided how we'll manage that.

src/lib/Repository/UiLocationService.php
```diff
@@ -52,6 +63,26 @@ public function loadLocations(ContentInfo $contentInfo)
...
+    public function swapLocations(int $locationId, int $newLocationId)
+    {
+        $currentLocation = $this->locationService->loadLocation($locationId);
+        $newLocation = $this->locationService->loadLocation($newLocationId);
+
+        $childCount = $this->locationService->getLocationChildCount($currentLocation);
+        $contentType = $this->contentTypeService->loadContentType($newLocation->getContentInfo()->contentTypeId);
+
+        if (!$contentType->isContainer && $childCount) {
+            throw new ForbiddenException('Cannot swap location that has sub items with a location that is not a container');
```
dpobel: `ForbiddenException` does not feel right to me. `InvalidArgumentException` maybe ?

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -39,3 +39,18 @@
     {% endif %}
     </tbody>
 </table>
+
+<div class="ez-locations-box">
```
dpobel: to get something visually closer to the PM definition, the markup should be something like:

```html
<form action="/whatever/uri" method="post">
<div class="ez-list-toolbar">
    <h2 class="ez-list-toolbar-label">Location content swap</h2>
</div>
<p>
    Swap the content item at this location with another
    <button type="submit" class="ez-button ez-button-ghost">Select content item</button>
    <input type="number" placeholder="Location ID">
</p>
</form>
```

dpobel (changes requested):

src/bundle/Resources/views/content/tabs/locations.html.twig
maxbaldanza: I didn't realise that was the case 👍 , was trying to follow the UI guidelines. Shall I remove the `SubmitTypeIconExtension.php` class too if this isn't needed?

bdunogier: In general, consider that any new markup, moreover css & js, should include a ping to Damien :)

src/bundle/Resources/views/form/fields.html.twig
```diff
@@ -0,0 +1,18 @@
+{%- block button_widget -%}
```
bdunogier: Wouldn't `{spaceless}` make the code easier to read ?

bdunogier:
Mostly good to go as far as I'm concerned, with @dpobel's remarks taken into account.

src/bundle/Resources/views/content/tabs/locations.html.twig
dpobel: @maxbaldanza I guess yes, it's better to remove the extension for now as I really don't know what this will look like with the icon :-/ (and I'm sorry for that)

maxbaldanza: Will do next time 👍 I'll remove it and put it on a separate branch in case it needs to come back at a later time

src/lib/Repository/UiLocationService.php
maxbaldanza: Either would work. Although I was thinking of creating a new exception for it to make it more specific something like `LocationSwapForbiddenException`. What are your thoughts on making the exceptions more specific or should we stick to the more generic ones?

src/bundle/Resources/views/content/tabs/locations.html.twig
maxbaldanza: Removed from this PR and added it to a separate branch in case it is needed in the future: https://github.com/inviqa/hybrid-platform-ui/tree/add_icon_to_symfony_form_buttons

src/bundle/Resources/config/routing.yml
```diff
@@ -18,3 +18,8 @@ ez_hybrid_platform_ui_location_actions:
     defaults:
         _controller: EzSystemsHybridPlatformUiBundle:Location:actions
     methods: [POST]
+ez_hybrid_platform_ui_location_swap_location:
+    path: /content/{contentId}/location/swap-location/{locationId}
```
bdunogier: I'd prefer `/content/{contentId}/location/{locationId}/swap`.

src/lib/Repository/UiLocationService.php
```diff
@@ -75,6 +84,29 @@ public function deleteLocations(array $locationIds)
         }
     }
 
+    /**
+     * Swaps locations.
+     *
+     * @param int $locationId
+     * @param int $newLocationId
+     */
+    public function swapLocations(int $locationId, int $newLocationId)
```
bdunogier: `mixed`, not `int`.

bdunogier:
@maxbaldanza the description says that it needs to be updated depending on the outcome of #44. Since it has been merged, I guess that something needs to be updated (description or code).

maxbaldanza:
Good spot @bdunogier I've now updated it to use `TabController::resetLocation` based on the discussion in #44

bdunogier (approved):

src/bundle/Controller/LocationController.php
```diff
@@ -83,6 +85,26 @@ public function actionsAction(
...
+        if ($swapLocationsForm->isValid()) {
+            $newLocationId = $swapLocationsForm->get('new_location_id')->getData();
+            $this->uiLocationService->swapLocations($location, $newLocationId);
+
+            return $this->resetLocation($content->id);
+        }
+
+        $redirectLocationId = $request->query->get('redirectLocationId', $content->contentInfo->mainLocationId);
+
+        return $this->reloadTab('locations', $content->id, $redirectLocationId);
```
bdunogier, line 105: I don't think `reloadTab()` will work here.

We are viewing `Location A`, referring to `Content X`, and we are swapping it with `Location B`, that refers to `Content Y`. After the swap, `Location A` refers to `Content Y`, and `Location B` refers to `Content X`.

If we only reload the tab, we are still viewing `Location A`, and only the Locations tab is reloaded. Most, if not all, of the other tabs are incorrect, since they still refer to `Content X`. We should call `resetLocation()`, either with the originally viewed location, or the one we swapped it for. People here said that the target location is the most logical, as it keeps you on the same content item.

Also, why do we need a `redirectLocationId` ? We know about both of the involved locations, and shouldn't need that.

bdunogier (changes requested):

src/bundle/Controller/LocationController.php
bdunogier: @SylvainGuittard has replied on JIRA. Please change this to a `resetLocation()` with the source location (`$location`).

maxbaldanza: `resetLocation` is used on line 100 when the form is valid. 
`reloadTab` will only be used if the form is invalid to take you back to the page you were on.

bdunogier: My bad. Okay, great :)

But it was specified by @SylvainGuittard that we should reset to the _same location id_, with the new content it has received. It kind of indicates that `resetLocation()` should be modified so that it can accept a Location, or location id (I'd prefer Location).

About it, actually, if it has to work with a `Content`, I'd create a distinct method, with a clear name like `resetToMainLocation($content)`.

maxbaldanza: I've updated the PR to now redirect back to the same location with the new content

bdunogier (approved):

PR #55, opened 2017-07-05
EZP-27577: style <ez-field-edit>

StephaneDiot:
JIRA: https://jira.ez.no/browse/EZP-27577
## Description

Adding a bit of CSS for <ez-field-edit> validation

### DIFF
--- a/src/bundle/Resources/public/css/modules/field-edit.css
+++ b/src/bundle/Resources/public/css/modules/field-edit.css
@@ -1,0 +1,19 @@
+ez-field-edit input:invalid {
+    border: 1px solid #d92d42;
+    background-color: #fceaec;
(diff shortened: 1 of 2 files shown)


src/bundle/Resources/public/css/modules/field-edit.css
```diff
@@ -0,0 +1,19 @@
+ez-field-edit input:invalid {
```
dpobel, line 1: `select` and `textarea` can also be invalid. That said, this kind of rule should also be applied to forms that don't involve fields (ie forms in the admin panel). So I think we should take some time to think about forms in general to not make the same mistake as in PlatformUI where forms were mostly approached for Content edit resulting in poor admin forms.

StephaneDiot:
Closed for now as we can think to the good approach

PR #57, opened 2017-07-06
EZP-27506 toggle locations

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27506

image

This PR introduces the following:

- Notification Builder
- Notification Response

although the primary purpose is to show/hide locations

some additional FE work will be required to style the checkbox and also to populate and submit the hidden visibility form with the current location id and visiilty state when an item is checked/unchecked. The notification response from that form indicates whether to toggle the state of the visible checkbox or not.

### DIFF
--- a/src/lib/Builder/NotificationBuilder.php
+++ b/src/lib/Builder/NotificationBuilder.php
@@ -1,0 +1,143 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Builder;
+
+use Twig\Environment;
+
+/**
+ * Builder for notification messages, returns output from twig.
+ */
+class NotificationBuilder
+{
+    const TEMPLATE = 'EzSystemsHybridPlatformUiBundle::notification.html.twig';
+
+    const DEFAULT_TIMEOUT = 10;
+
+    const TYPE_SUCCESS = 'positive';
+
+    const TYPE_ERROR = 'error';
+
+    /**
+     * @var Environment
+     */
+    private $environment;
+
+    /**
+     * @var string
+     */
+    private $type;
+
+    /**
+     * @var string
+     */
+    private $message;
+
+    /**
+     * @var int
+     */
+    private $timeout = self::DEFAULT_TIMEOUT;
+
+    /**
+     * @var bool
+     */
+    private $copyable;
+
+    /**
+     * @var string
+     */
+    private $details;
+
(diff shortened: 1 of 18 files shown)


src/lib/Builder/NotificationBuilder.php
```diff
@@ -0,0 +1,143 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Builder;
+
+use Twig\Environment;
+
+/**
+ * Builder for notification messages, returns output from twig.
+ */
+class NotificationBuilder
```
dpobel, line 13: hum @bdunogier while working on https://jira.ez.no/browse/EZP-27476 will also need to handle Notification and we were also talking about that lately in https://github.com/ezsystems/hybrid-platform-ui/pull/30#discussion_r124520732 where we were thinking about a `Notification` value object.
So that's a hot topic where we (you :p) need to synchronize

src/lib/Response/NotificationResponse.php
```diff
@@ -0,0 +1,38 @@
...
+    /**
+     * Error response.
+     *
+     * @param string $content
+     *
+     * @return static
+     */
+    public static function error(string $content)
+    {
+        return new static($content, static::HTTP_BAD_REQUEST);
```
dpobel, line 36: I think it's a bad idea to _restrict_ the HTTP status code here. Because of that we would loose information on why the request failed and those information might be critical for the frontend app to correctly handle some situations.

src/lib/EventSubscriber/AppRendererSubscriber.php
```diff
@@ -51,7 +52,7 @@ public function renderApp(FilterResponseEvent $event)
 
         $response = $event->getResponse();
 
-        if ($response->isRedirect()) {
+        if ($response->isRedirect() || $response instanceof NotificationResponse) {
```
dpobel, line 55: unsure why you need to do that ? isn't it because https://jira.ez.no/browse/EZP-27582 / https://jira.ez.no/browse/EZP-27476 are not yet implemented ?

dpobel (changes requested):

src/lib/Response/NotificationResponse.php
mcurran-inviqa: a notification will be included to tell you why it failed

src/lib/EventSubscriber/AppRendererSubscriber.php
mcurran-inviqa: it's because the app renderer kicks in unless we tell it not to

PR #58, opened 2017-07-06
EZP-27503 add location

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27503

image

This PR adds a the basic mechanic of creating a location, currently this is done via manually entering a location id.

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -69,14 +69,28 @@ class LocationController extends TabController
         $redirectLocationId = $request->query->get('redirectLocationId', $content->contentInfo->mainLocationId);
 
         if ($actionsForm->isValid()) {
-            $this->deleteLocationsBasedOnFormSubmit($actionsForm, $content, $redirectLocationId);
+            $resetLocation = $this->deleteLocationsBasedOnFormSubmit($actionsForm, $redirectLocationId);
+
+            if ($resetLocation) {
+                return $this->resetLocation($content->id);
+            }
+
             $this->addLocationBasedOnFormSubmit($actionsForm, $content);
         }
 
         return $this->reloadTab('locations', $content->id, $redirectLocationId);
     }
 
-    private function deleteLocationsBasedOnFormSubmit(FormInterface $form, Content $content, $redirectLocationId)
+    /**
+     * Delete locations based on form submit.
+     *
+     * @param FormInterface $form
+     * @param mixed $redirectLocationId
+     *
+     * @return bool
+     *  Whether to reset location or not.
+     */
+    private function deleteLocationsBasedOnFormSubmit(FormInterface $form, $redirectLocationId)
     {
         $locationIds = array_keys($form->get('removeLocations')->getData());
 
@@ -84,9 +98,11 @@ class LocationController extends TabController
             $this->uiLocationService->deleteLocations($locationIds);
 
             if (in_array($redirectLocationId, $locationIds)) {
-                return $this->resetLocation($content->id);
+                return true;
             }
         }
+
+        return false;
     }
 
     private function addLocationBasedOnFormSubmit(FormInterface $form, Content $content)
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -8,25 +8,47 @@
  */
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
+use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
 use EzSystems\HybridPlatformUi\Repository\UiLocationService;
+use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Routing\RouterInterface;
 
 class LocationController extends TabController
 {
-    public function contentViewTabAction(
-        ContentView $view,
+    /**
+     * @var UiLocationService
+     */
+    private $uiLocationService;
+
+    /**
+     * @var UiFormFactory
+     */
+    private $formFactory;
+
+    public function __construct(
         UiLocationService $uiLocationService,
+        RouterInterface $router,
+        ContentService $contentService,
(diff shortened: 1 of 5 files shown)


src/bundle/Form/Locations/Actions.php
```diff
@@ -17,6 +18,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
             ->add('delete', SubmitType::class)
+            ->add('add', SubmitType::class)
+            ->add('parentLocationId', HiddenType::class)
```
dpobel: for now it's can even be a TextType, that way you can test it manually (it's specified like that  in https://jira.ez.no/browse/EZP-27503 see attachment)

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -4,6 +4,7 @@
 <div class="ez-list-toolbar">
     <h2 class="ez-list-toolbar-label">{{ 'locationview.locations.content.locations'|trans()|desc('Content locations') }}</h2>
     <div class="container-buttons">
+        {{ form_widget(actionsForm.add, {'label': 'locationview.locations.add.location'|trans()|desc('Add Location'), 'attr': {'class': 'ez-button ez-button-secondary'}}) }}
```
dpobel: so following my previous comment, you should also render `actionsForm.parentLocationId` here

src/bundle/Controller/LocationController.php
```diff
@@ -56,6 +56,13 @@ public function actionsAction(
             if ($actionsForm->get('delete')->isClicked()) {
                 $uiLocationService->deleteLocations($locationIds);
             }
+
+            if ($actionsForm->get('add')->isClicked()) {
+                $uiLocationService->addLocation(
+                    $content->contentInfo,
+                    $actionsForm->get('parentLocationId')->getData()
+                );
+            }
```
dpobel: and btw, it also a good occasion to test the new forms in tabs feature, so you should:

1. rebase on master
1. make sure to redirect the `locations_tab` view below

dpobel (changes requested):

src/lib/Repository/UiLocationService.php
```diff
@@ -75,6 +75,18 @@ public function deleteLocations(array $locationIds)
         }
     }
 
+    /**
+     * Creates location.
+     *
+     * @param ContentInfo $contentInfo
+     * @param $parentLocationId
```
maxbaldanza: Can we add the type to `$parentLocationId`

src/bundle/Form/Locations/Actions.php
```diff
@@ -17,6 +18,8 @@ public function buildForm(FormBuilderInterface $builder, array $options)
     {
         $builder
             ->add('delete', SubmitType::class)
+            ->add('add', SubmitType::class)
+            ->add('parentLocationId', TextType::class)
```
maxbaldanza: As @bdunogier mentioned in the stand up could we not set the `data_class` of this to be `LocationCreateStruct` and this should avoid the call to `LocationService::newLocationCreateStruct`

src/bundle/Controller/LocationController.php
```diff
@@ -58,6 +58,13 @@ public function actionsAction(
                     return $this->resetLocation($content->id);
                 }
             }
+
+            if ($actionsForm->get('add')->isClicked()) {
```
maxbaldanza: Would it be worth refactoring this and the deletion of location to a private method with a descriptive name to make it easier to read?

maxbaldanza (changes requested):

src/bundle/Form/Locations/Actions.php
mcurran-inviqa: we can't use data_class for this form, it's multi purpose

src/bundle/Controller/LocationController.php
```diff
@@ -8,25 +8,47 @@
...
+
+    /**
+     * @var UiFormFactory
+     */
+    private $formFactory;
+
+    public function __construct(
         UiLocationService $uiLocationService,
+        RouterInterface $router,
+        ContentService $contentService,
```
maxbaldanza: Probably one for a different PR but I'm wondering if `TabController` can be refactored to remove the dependency on ContentService? Seems strange that we need `ContentService` in this controller especially when resetLocation isn't even called

maxbaldanza (approved):

src/bundle/Controller/LocationController.php
```diff
@@ -81,4 +99,38 @@ public function updateDefaultSortOrderAction(
...
         return $this->reloadTab('details', $content->id, $location->id);
     }
+
+    /**
+     * Delete locations based on form submit.
+     *
+     * @param FormInterface $form
+     * @param mixed $redirectLocationId
+     *
+     * @return bool
```
bdunogier: Needs to be on the same line than `@return bool`.

src/bundle/Controller/LocationController.php
```diff
@@ -81,4 +99,38 @@ public function updateDefaultSortOrderAction(
...
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    private function addLocationBasedOnFormSubmit(FormInterface $form, Content $content)
+    {
+        if ($form->get('add')->isClicked()) {
```
bdunogier, line 128: About this test, and the equivalent one in the `deleteLocations` method above, don't you think that it would make more sense to test for that in `actionsAction` ?

bdunogier:
Besides the two minor remarks, we're almost good.

src/bundle/Controller/LocationController.php
mcurran-inviqa: it's in line with how it's done on the versions tab, removes some clutter from the actionsAction method

maxbaldanza: I suggested this change as I think it makes the action itself easier to read

bdunogier: Well, there is clutter, and there is understandability. The code always ends up somewhere :)

Not a blocker for this merge, but I personally think that it would make more sense to test in the action, and process the action that was tested positive. We call the process methods N times while only 1 can be performed at a time, it's honestly a bit weird :)

About versions using it as well, it's not a magical argument: things can be consistently "bad" ;)

PR #59, opened 2017-07-06
EZP-27447 display related content

maxbaldanza:
- [x] Do not show reverse relations table if the user doesn't have access.
- [x] Add edit button to table
- [x] Update labels

https://jira.ez.no/browse/EZP-27447

### DIFF
--- a/src/lib/Repository/UiRelationService.php
+++ b/src/lib/Repository/UiRelationService.php
@@ -62,17 +62,18 @@ class UiRelationService
     /**
      * Loads reverse relations for a piece of content.
      * Retrieves reverse relations and then sets the field definition and the content type name.
+     * Returns false if the user doesn't have permission to load reverse relations.
      *
      * @param ContentInfo $contentInfo
      *
-     * @return UiRelation[]
+     * @return UiRelation[]|bool
      */
     public function loadReverseRelations(ContentInfo $contentInfo)
     {
         try {
             $reverseRelations = $this->contentService->loadReverseRelations($contentInfo);
         } catch (UnauthorizedException $e) {
-            return [];
+            return false;
         }
 
         return $this->buildSourceUiRelations($reverseRelations);
--- a/src/lib/Repository/UiRelationService.php
+++ b/src/lib/Repository/UiRelationService.php
@@ -7,7 +7,6 @@ namespace EzSystems\HybridPlatformUi\Repository;
 
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\API\Repository\ContentTypeService;
-use eZ\Publish\API\Repository\Exceptions\UnauthorizedException;
 use eZ\Publish\API\Repository\LocationService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\API\Repository\Values\Content\Relation;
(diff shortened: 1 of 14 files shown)


src/lib/Repository/UiRelationService.php
```diff
@@ -62,17 +62,18 @@ public function loadRelations(VersionInfo $versionInfo)
...
      * @param ContentInfo $contentInfo
      *
-     * @return UiRelation[]
+     * @return UiRelation[]|bool
      */
     public function loadReverseRelations(ContentInfo $contentInfo)
     {
         try {
             $reverseRelations = $this->contentService->loadReverseRelations($contentInfo);
         } catch (UnauthorizedException $e) {
```
bdunogier: I'd prefer if we tested the permissions instead of trying and catching. Exceptions should be kept for exceptional situations.

bdunogier (changes requested):
Besides the small remark about exception handling, it looks good to me.

src/lib/Repository/UiRelationService.php
maxbaldanza: The reason it was implemented like this is is to avoid duplicating the code that checks the permissions. So that permissions can remain as part of the kernel and therefore would have one central place where they would need to be edited. 
Happy to add to here if you'd prefer though?

bdunogier: I'd really prefer if we tested explicitly.

In addition, a good remark from @dpobel : `false` is a bit of a magic value. Wrapping it as a constant would make sense here, like `UiRelationService::NO_REVERSE_RELATIONS_PERMISSION` (set to `false`).

Then we can either test it in the template (I find the syntax quite ugly), or test it in the controller, and assign a specific variable instead of `reverseRelations` that we test instead of `same as (false)`.

maxbaldanza: I'm not sure using a constant is a much better approach. I've updated the PR to check the permission beforehand which I believe is a cleaner approach. Let me know what you think?

src/lib/Repository/UiRelationService.php
```diff
@@ -62,19 +61,16 @@ public function loadRelations(VersionInfo $versionInfo)
...
+     * @throws \eZ\Publish\API\Repository\Exceptions\UnauthorizedException if the user is not allowed to read reverse relations.
      *
      * @param ContentInfo $contentInfo
      *
-     * @return UiRelation[]|bool
+     * @return UiRelation[]
      */
     public function loadReverseRelations(ContentInfo $contentInfo)
     {
-        try {
```
bdunogier: Hmm, I may not have explained correctly. It is good that you're using the resolver, but I was suggesting that we check for permissions _inside this method_ instead of in the action.

As said in another comment, if we consider that those services implement the business logic of the hybrid UI, it makes sense.

bdunogier:
Almost there :)

src/lib/Repository/UiRelationService.php
maxbaldanza: I don't like the idea of returning a constant or an array from within the method as it's not expected behaviour. I think keeping this generic without checking for permissions may be better as there may come a time when we need the reverse relations and it should throw an exception if someone doesn't have access. 
Thoughts?

mcurran-inviqa (approved):

src/lib/View/Content/ParameterSupplier.php
```diff
@@ -0,0 +1,22 @@
...
+/**
+* @copyright Copyright (C) eZ Systems AS. All rights reserved.
+* @license For full copyright and license information view LICENSE file distributed with this source code.
+*/
+namespace EzSystems\HybridPlatformUi\View\Content;
+
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+
+/**
+ * Supplies the controller with any parameters it may require to display the view.
```
bdunogier: It actually supplies the `View` with parameters, not the controller.

bdunogier (changes requested):

bdunogier:
@maxbaldanza unless the commits list gets squashed in a more granular way by you, maybe separating the Supplier's addition from the feature itself, I'm gonna squash the whole thing into one commit.

Any preference ?

maxbaldanza:
No preference either way, I'm happy for it all be squashed in to one commit as it's all part of the same story.

PR #61, opened 2017-07-06
EZP-27610 location redirect

mcurran-inviqa:
https://jira.ez.no/browse/EZP-27610

### DIFF
--- a/src/bundle/Controller/TabController.php
+++ b/src/bundle/Controller/TabController.php
@@ -1,0 +1,71 @@
+<?php
+/**
+ * File containing the TabController class.
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUiBundle\Controller;
+
+use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
+use EzSystems\HybridPlatformUi\Response\ResetResponse;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\Routing\RouterInterface;
+
+abstract class TabController extends Controller
+{
+    /**
+     * @var RouterInterface
+     */
+    private $router;
+
+    /**
+     * @var ContentService
+     */
+    private $contentService;
+
+    public function __construct(
+        RouterInterface $router,
+        ContentService $contentService
+    ) {
+        $this->router = $router;
+        $this->contentService = $contentService;
+    }
+
+    protected function resetLocation($contentId)
+    {
+        $mainLocationId = $this->contentService->loadContentInfo($contentId)->mainLocationId;
+
+        return new ResetResponse($this->generateTabUrl($contentId, $mainLocationId));
+    }
+
+    protected function redirectToLocationsTab($contentId, $locationId)
+    {
+        return $this->redirectToTab($contentId, $locationId, 'locations_tab');
+    }
+
+    protected function redirectToVersionsTab($contentId, $locationId)
+    {
+        return $this->redirectToTab($contentId, $locationId, 'versions_tab');
+    }
+
+    private function redirectToTab($contentId, $locationId, $viewType)
+    {
+        return new RedirectResponse(
+            $this->generateTabUrl($contentId, $locationId, $viewType)
+        );
+    }
+
+    private function generateTabUrl($contentId, $locationId, $viewType = 'full')
+    {
+        return $this->router->generate(
+            '_ez_content_view',
+            [
+                'contentId' => $contentId,
+                'locationId' => $locationId,
+                'viewType' => $viewType,
+            ]
+        );
+    }
+}
--- a/src/bundle/Controller/TabController.php
+++ b/src/bundle/Controller/TabController.php
@@ -18,12 +18,12 @@ abstract class TabController extends Controller
     /**
      * @var RouterInterface
      */
-    private $router;
+    protected $router;
 
     /**
      * @var ContentService
      */
-    private $contentService;
+    protected $contentService;
 
     public function __construct(
         RouterInterface $router,
--- a/src/bundle/Controller/TabController.php
+++ b/src/bundle/Controller/TabController.php
@@ -9,7 +9,7 @@ namespace EzSystems\HybridPlatformUiBundle\Controller;
 
 use eZ\Publish\API\Repository\ContentService;
 use eZ\Publish\Core\MVC\Symfony\Controller\Controller;
-use EzSystems\HybridPlatformUi\Response\ResetResponse;
+use EzSystems\HybridPlatformUi\Http\Response\ResetResponse;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\Routing\RouterInterface;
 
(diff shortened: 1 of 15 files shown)


src/bundle/Controller/TabController.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    private function generateTabUrl($contentId, $locationId, $viewType = 'full')
+    {
+        return $this->router->generate(
+            '_ez_content_view',
+            [
+                'contentId' => $contentId,
+                'locationId' => $locationId,
+                'viewType' => $viewType,
```
bdunogier, line 62: This would be a good spot to suffix with `_tab`.

src/lib/EventSubscriber/AppRendererSubscriber.php
```diff
@@ -51,7 +51,7 @@ public function renderApp(FilterResponseEvent $event)
 
         $response = $event->getResponse();
 
-        if ($response->isRedirect()) {
+        if ($response->isRedirect() || $response instanceof NoRenderResponse) {
```
bdunogier, line 54: Or you could implement `isRedirect` in `ResetResponse`. It has a `Location` header (or close enough), I think that's clean.

src/lib/Response/NoRenderResponse.php
```diff
@@ -0,0 +1,13 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Response;
+
+/**
+ * Interface to indicate that this response should bypass the app renderer.
```
bdunogier: With `isRedirect()` implemented as suggested above that interface isn't necessary anymore.

src/lib/Response/ResetResponse.php
```diff
@@ -0,0 +1,25 @@
...
+namespace EzSystems\HybridPlatformUi\Response;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Reset response to tell the app to reload the location.
+ */
+class ResetResponse extends Response implements NoRenderResponse
+{
+    public function __construct(string $url = '')
```
bdunogier: I'd say that the URL is required in our case, wouldn't you ? We don't need a default value.

bdunogier (changes requested):

src/lib/EventSubscriber/AppRendererSubscriber.php
mcurran-inviqa: there will be other Response classes that use this, isRedirect is in the parent and checks for the 3xx codes, I don't like to override it and class something that isn't technically a redirect as a redirect.

src/lib/Response/ResetResponse.php
mcurran-inviqa: This is because of a problem with autowiring it tries to inject a parameter unless a default is specified.

src/bundle/Controller/TabController.php
mcurran-inviqa: not all view types have _tab, for instance full

src/bundle/Controller/TabController.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    protected function resetLocation($contentId)
+    {
+        $mainLocationId = $this->contentService->loadContentInfo($contentId)->mainLocationId;
+
+        return new ResetResponse($this->generateTabUrl($contentId, $mainLocationId));
+    }
+
+    protected function redirectToLocationsTab($contentId, $locationId)
```
bdunogier: I wouldn't do that.

Tabs are meant to be extensible in the end, this won't be usable. Instead, I'd assume that we standardize the name for the tabs views (`xxx_tab`). We can pass the tab identifier to the method, and `_tab` will be added automatically:

```
$this->redirectToTab('locations', $contentId, $locationId);
```

What bothers me a bit more is that in that case, it is NOT a redirection, from the hybrid-platform-ui perspective. The protocol we have established makes that a refresh, hence my suggestion in the story.

src/lib/Response/NoRenderResponse.php
```diff
@@ -0,0 +1,13 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Response;
```
bdunogier: I'd move all Response classes to the `Http` directory.

src/lib/EventSubscriber/AppRendererSubscriber.php
bdunogier: > There will be other Response classes that use this
Which ones ?

I don't see the problem with it being in the parent. That is what inheritance is for.
The App doesn't render the App structure when the Response is a redirect. The `ResetResponse` acts like a redirect, but isn't really one because of frontend requirements. 

I'd like to insist on that test :)

bdunogier (changes requested):

src/lib/Response/ResetResponse.php
```diff
@@ -0,0 +1,25 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Response;
+
+use Symfony\Component\HttpFoundation\Response;
+
+/**
+ * Reset response to tell the app to reload the location.
+ */
+class ResetResponse extends Response implements NoRenderResponse
```
bdunogier: Wouldn't that be `ResetLocationResponse` ? The `App-Location` header makes it quite specific. But do we need it to be location specific @dpobel ? We may want to redirect from a tab to an entirely different part of the admin.

src/bundle/Controller/TabController.php
dpobel: if you redirect to the `full` view in a method called `generateTabUrl` so dedicated to tabs, something is wrong somewhere :)

src/lib/Response/ResetResponse.php
dpobel: > We may want to redirect from a tab to an entirely different part of the admin.

that's the point of this kind of response. But maybe there's a confusion because of `Location` here? `App-Location` is inspired by `Location` HTTP header from redirect responses, so `Location` here is unrelated to our Location object, it can contain any URI of the admin

dpobel (changes requested):

src/bundle/Controller/TabController.php
bdunogier: > not all view types have _tab, for instance full

This is the *tabController*, it is not meant to use the `full` viewtype, or any other viewtype that isn't a tab.

src/bundle/Controller/TabController.php
```diff
@@ -0,0 +1,66 @@
...
+
+        return new ResetResponse($this->generateContentViewUrl($contentId, $mainLocationId));
+    }
+
+    protected function reloadTab($tab, $contentId, $locationId)
+    {
+        return $this->redirectToTab($contentId, $locationId, $tab . '_tab');
+    }
+
+    private function redirectToTab($contentId, $locationId, $viewType)
```
bdunogier, line 48: This feels like it duplicates `reloadTab`, and the `RedirectResponse` could be instanciated inside that method instead.

src/bundle/Controller/TabController.php
bdunogier: @mcurran-inviqa is correct, the `_tab` suffix was added to `reloadTab()`, and it is the right place. This private method is also used by `resetLocation()` to generate the `full` view url.

src/lib/EventSubscriber/AppRendererSubscriber.php
bdunogier: Let's keep it as is for the time being. Changes are necessary in that area anyway.

bdunogier (approved):

maxbaldanza (approved):

src/bundle/Controller/TabController.php
dpobel: ok :)

dpobel (approved):

PR #67, opened 2017-07-07
EZP-27617 content tabs notification

mcurran-inviqa:
- Adds a NotificationMessage value object
- Adds NotificationBuilder for building NotificationMessage
- Adds a factory for NotificationBuilder to allow multiple Notifications in the same request
- Adds a custom response type for Notifications to bypass app rendering

https://jira.ez.no/browse/EZP-27617

based on the ideas found in https://gist.github.com/dpobel/875e9655f05e0eaeb23c795f6f3dfe47


src/lib/Http/Response/NotificationResponse.php
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Error response.
+     *
+     * @param string $content
+     * @param int $status
+     *
+     * @return static
+     */
+    public static function error(string $content, int $status)
```
mcurran-inviqa: it may make for a better interface if we have limited errors to have a method for each one instead of passing in a status thoughts?

src/lib/Builder/NotificationBuilder.php
```diff
@@ -0,0 +1,143 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Builds notification message.
+     *
+     * @return string
+     */
+    public function getResult()
```
dpobel: you are coupling the Notification creation and rendering and I think it's a mistake for 2 reasons:

1. Notifications won't always be rendered the same way. We already render some notifications in Hybrid Platform UI (when coming from PJAX response handled by PlatformUIBundle) and those are set on the app [to be serialized in JSON](https://github.com/ezsystems/hybrid-platform-ui/blob/master/src/lib/Components/App.php#L85). @bdunogier is also setting up pretty much the same [in the PR about error handling](https://github.com/ezsystems/hybrid-platform-ui/pull/68/files#diff-d4bd0f0dbb343592a25da1345613ae69R70).
2. The creation of a Notification and its rendering is usually done very separately. For instance, after a form submit, an action is done which ends up with one or several notifications, then you are redirected and after that notifications are rendered.

So for me, we should first have a `Notification` value object (which could be extended to have for instance an `ErrorNotification` where `timeout` is forced to 0) to represent the concept of Notification. Then how those are rendered depends on where there are used.

For tabs, I think each tab should store its own notifications in its own entry in the flash bag. So the default operation flow in tabs should be the following:

1. the user submits a form in a tab
2. the form is handled, one or several notifications are put in a flash bag entry dedicated to the tab, then a redirection to the tab action is generated
3. when rendering the tab, the tab flash bag entry is read to render the `ez-notification` elements along the tab content

Now remains the case of Show/Hide Location (and possibly for set main operation) where there should be no redirection to the tab view. For those, I guess after processing the form, the redirection could be made to a very simple action that would respond with an HTTP status corresponding to what happened while processing the form and would just read the corresponding flash bag.

dpobel (changes requested):

bdunogier:
Ping @dpobel. New review needed.

src/lib/Builder/NotificationBuilder.php
mcurran-inviqa: this PR was updated not long after this comment, it is for the show/hide response it introduces generic notification value objects which now have a toString method for rendering in a response, they are passed to the response class and that will automatically call toString on them.

we should probably be unifying the display of notifications regardless of how they got onto to the page so will need some further discussion.

bdunogier: Notifications aren't only meant to be displayed as HTML. They're also rendered to json for update responses.

bdunogier: Anyway, I'm overall not convinced.

The builder doing the rendering doesn't work very well, as explained by Damien and myself. HTML rendering of a notification is only one aspect of it.

In addition, we need to remember that notifications are _mostly_ meant to be passed to the flashbag, so that the app can pick them up during rendering. The builder and the factory are meant to create the object, but do not handle the flashbag. It means that developers using the feature will have to inject both services to create a notification. I'd prefer if we had one service that is used to _add the notification_.

mcurran-inviqa: the Builder does not render it just creates a VO, the VO does handle rendering.

src/lib/Notification/Notification.php
```diff
@@ -0,0 +1,66 @@
+<?php
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUi\Notification;
+
+use eZ\Publish\API\Repository\Values\ValueObject;
+
```
maxbaldanza, line 9: Can we add doc block explaining what it does?

maxbaldanza (changes requested):

src/lib/Http/Session/Flash/NotificationBag.php
```diff
@@ -0,0 +1,96 @@
...
+
+use EzSystems\HybridPlatformUi\Notification\Notification;
+use Symfony\Component\HttpFoundation\Session\Flash\FlashBagInterface;
+use Symfony\Component\HttpFoundation\Session\Session;
+use Symfony\Component\Translation\TranslatorInterface;
+
+/**
+ * Used for adding notifications.
+ */
+class NotificationBag
```
bdunogier: This needs an interface, so that we have the option to swap it with another implementation.

We could use `NotificationBag` as the interface, and `SessionNotificationBag` as the implementation. Naming the interface without "bag" would also make sense if you find something.

src/lib/Http/Session/Flash/NotificationBag.php
```diff
@@ -0,0 +1,96 @@
...
+{
+    const FLASH_MESSAGE_TYPE = 'hyybrid-notification';
```
bdunogier: @dpobel did we not say that we wanted to have notifications per element ? Like tabs notifications ?

src/lib/Notification/Notification.php
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Notification object for messages to users, to be stored in a FlashBag.
```
bdunogier: I wouldn't mention the FlashBag here. The interface maybe.

src/lib/Notification/Notification.php
```diff
@@ -0,0 +1,76 @@
...
+    protected $copyable;
+
+    /**
+     * Extended details for Developers.
+     *
+     * @var string
+     */
+    protected $details;
+
+    public function __toString()
```
bdunogier: I'm not sure we should keep that.

The rendering of a notification to a string can very well be done, using `sprintf` as well, from the `NotificationResponse`.

bdunogier (changes requested):
Looks mostly good, just a couple remarks.

Will need to be part of a developer documentation.

src/lib/Http/Session/Flash/NotificationBag.php
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    /**
+     * Add success notification.
+     *
+     * @param string $message
+     * @param array $params
+     * @param null $domain
+     */
+    public function addSuccess(string $message, array $params = [], $domain = null)
```
bdunogier: I think that we should have those methods available in the `TabController`, as abstracts. We may even want to _move_ the `Notification` object building to these abstracts. That way, we don't tie the type of notifications we support to this object, meaning that we can easily add more, including in 3rd party code.

Could even be part of a Trait that the abstract controller uses ?

src/lib/Http/Response/NotificationResponse.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    public function __construct(Notification $notification, $status = Response::HTTP_OK)
+    {
+        parent::__construct($this->convertNotificationToString($notification), $status);
+    }
+
+    private function convertNotificationToString(Notification $notification)
+    {
+        return sprintf(
+            '<ez-notification type="%s" timeout="%s"%s%s><p>%s</p></ez-notification>',
```
dpobel: I think it's a mistake to force a paragraph here. A notification message can be in any markup.

src/lib/Notification/Notification.php
```diff
@@ -0,0 +1,64 @@
...
+     * @var string
+     */
+    protected $message;
+
+    /**
+     * Message timeout.
+     *
+     * @var int
+     */
+    protected $timeout;
```
dpobel: you should set the default value here

src/lib/Notification/NotificationPoolAwareTrait.php
```diff
@@ -0,0 +1,86 @@
...
+     * @param array $params
+     * @param string|null $domain
+     */
+    protected function addSuccessNotification(string $message, array $params = [], $domain = null)
+    {
+        $notification = new Notification([
+            'type' => Notification::TYPE_SUCCESS,
+            /** @Ignore */
+            'message' => $this->translator->trans($message, $params, $domain),
+            'timeout' => Notification::DEFAULT_TIMEOUT,
```
dpobel: with the default value set in the value object, you don't have to set it here

dpobel (changes requested):

maxbaldanza (approved):

dpobel (approved):

bdunogier (approved):

PR #68, opened 2017-07-07
EZP-27476: app symfony error handling

bdunogier:
> Status: ready for review
> [EZP-27476](https://jira.ez.no/browse/EZP-27476)

Adds exception handling to the `AppRendererSubscriber`. On exceptions, an `App\ExceptionConfigurator` will configure the newly added App's `exception` configuration property. Rendering of the `App` is modified to render _only_ the exception when there is an error notification.

A new `Notifications` component is added. It uses the `NotificationPool` to fetch notifications upon rendering. If an exception is set through the app, it renders that exception alone instead of regular notifications.

In all components, `__toString()` is replaced by `renderToString()`, since exceptions can't be handled with it.

To facilitate testing, a refactoring of toolbars configuration has been done. They are now configured through the `AppRendererSubscriber`, using a `ToolbarsConfigurator` service. The current implementation uses a hardcoded route comparison, and enables the `discovery` bar for `ez_urlalias`.

It also refactors the app rendering by removing the `AppResponseRenderer`. It had always felt odd, as the App has its own interface: it can be converted to a string or serialized to json, and it is all we need. Doing so avoids the leak of Symfony specific objects (`Request`, `Response`) to the library layer. The downside is that it makes the renderer a _bit_ more complex, as it uses three request matchers. But the responsibilities feel like they're in the right place.

### TODO
- [x] Specs (`Components\Notifications`)
- [x] Finalize full page rendering of exceptions (401, 403, 404)
- [x] Verify that the app exception rendering based on 1 error notification is sufficient and doesn't have unwanted side effects
- [x] Remove test commit
- [x] Screenshots
- [x] Use Notification objects once merged

### DIFF
--- a/src/lib/App/RequestAppResponseRenderer.php
+++ b/src/lib/App/RequestAppResponseRenderer.php
@@ -8,15 +8,15 @@ namespace EzSystems\HybridPlatformUi\App;
 use EzSystems\HybridPlatformUi\Components\App;
 use EzSystems\HybridPlatformUi\Http\AjaxUpdateRequestMatcher;
 use Symfony\Component\HttpFoundation\JsonResponse;
-use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpFoundation\Response;
 
 class RequestAppResponseRenderer implements AppResponseRenderer
 {
     /**
-     * @var \Symfony\Component\HttpFoundation\Request
+     * @var \Symfony\Component\HttpFoundation\RequestStack
      */
-    private $request;
+    private $requestStack;
 
     /**
      * @var \Symfony\Component\HttpFoundation\RequestMatcherInterface
@@ -24,10 +24,10 @@ class RequestAppResponseRenderer implements AppResponseRenderer
     private $ajaxUpdateRequestMatcher;
 
     public function __construct(
-        Request $request,
+        RequestStack $requestStack,
         AjaxUpdateRequestMatcher $ajaxUpdateRequestMatcher
     ) {
-        $this->request = $request;
+        $this->requestStack = $requestStack;
         $this->ajaxUpdateRequestMatcher = $ajaxUpdateRequestMatcher;
     }
 
@@ -35,7 +35,7 @@ class RequestAppResponseRenderer implements AppResponseRenderer
     {
         $this->configureToolbars($app);
 
-        $appResponse = $this->ajaxUpdateRequestMatcher->matches($this->request)
+        $appResponse = $this->ajaxUpdateRequestMatcher->matches($this->requestStack->getCurrentRequest())
             ? new JsonResponse($app)
             : new Response($app->renderToString());
 
--- a/src/lib/App/RequestAppResponseRenderer.php
+++ b/src/lib/App/RequestAppResponseRenderer.php
@@ -35,13 +35,20 @@ class RequestAppResponseRenderer implements AppResponseRenderer
     {
         $this->configureToolbars($app);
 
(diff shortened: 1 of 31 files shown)


src/lib/App/RequestAppResponseRenderer.php
```diff
@@ -5,52 +5,76 @@
...
+        return $this->ajaxUpdateRequestMatcher->matches(
+            $this->requestStack->getMasterRequest()
+        );
+    }
+
+    private function buildNotification(Exception $exception)
     {
-        $app->setConfig(['toolbars' => ['discovery' => 1]]);
+        return [
+            'type' => 'error',
```
dpobel: I think we need a `Notification` value object (I know this collides with #67)

src/lib/Components/Component.php
```diff
@@ -5,4 +5,6 @@
 interface Component extends \JsonSerializable
 {
     public function __toString();
```
dpobel: I think you can remove `__toString` from this interface and remove `__toString` implementation in components implementing this interface

src/lib/Components/NavigationHub.php
```diff
@@ -33,15 +33,20 @@ class NavigationHub implements Component
      */
     protected $request;
 
-    public function __construct(EngineInterface $templating, Request $request, array $zones = [], array $links = [])
+    public function __construct(EngineInterface $templating, Request $request = null, array $zones = [], array $links = [])
```
dpobel, line 36: is there a reason to add a default value to `$request` ?

src/lib/Components/App.php
```diff
@@ -88,7 +94,18 @@ public function jsonSerialize()
...
+
+        return [
+            'selector' => self::TAG_NAME,
+            'update' => $update,
         ];
     }
+
+    private function isError()
+    {
+        return count($this->notifications) === 1
```
dpobel: that does not seem super reliable :) Maybe you can set a property on the app in `RequestAppResponseRenderer::configureAppException`

dpobel (changes requested):

src/lib/Components/App.php
bdunogier: The todo list for the PR agrees with you: _Verify that the app exception rendering based on 1 error notification is sufficient and doesn't have unwanted side effects_ :)

As said this morning, I was even considering using a different app option. These exceptions that are handled here are kind of peculiar.

src/lib/Components/Component.php
bdunogier: Indeed.

src/lib/Components/NavigationHub.php
bdunogier: Yes, `Fixed an error with Request when using debug:event-dispatcher`. CLI doesn't like the Request as a required service ctor argument.

src/lib/Components/App.php
bdunogier: Fixed now. Added an `exception` property to the app, that if is set will trigger the exception rendering.

src/lib/Components/Component.php
bdunogier: Done.

src/lib/App/RequestAppResponseRenderer.php
bdunogier: Indeed.

dpobel (approved):
I like the idea of the Notifications component :)

bdunogier:
> I like the idea of the Notifications component :)

Well, it was the clean alternative to having the component render itself. I'm still not a big fan of the components being Injectables as well as receiving values, but at least the Notifications component is consistently wrong ;)

PR #69, opened 2017-07-10
EZP-27568: Create draft from an archived version.

maxbaldanza:
- [x] Create draft from an archived version.
- [x] Validate form so that only one version can be selected when creating a new draft. Any number of versions can be selected when deleting.

https://jira.ez.no/browse/EZP-27568

### DIFF
--- a/src/bundle/Controller/VersionController.php
+++ b/src/bundle/Controller/VersionController.php
@@ -85,6 +85,7 @@ class VersionController extends TabController
 
         if ($archiveActionsForm->isValid()) {
             $this->deleteVersionsBasedOnFormSubmit($archiveActionsForm, $content->id);
+            $this->createDraftVersionBasedOnFormSubmit($archiveActionsForm, $content->id);
         }
 
         $redirectLocationId = $request->query->get('redirectLocationId', $content->contentInfo->mainLocationId);
@@ -100,6 +101,14 @@ class VersionController extends TabController
             foreach ($selectedIds as $versionId) {
                 $this->uiVersionService->deleteVersion((int) $contentId, $versionId);
             }
+        }
+    }
+
+    private function createDraftVersionBasedOnFormSubmit(FormInterface $form, $contentId)
+    {
+        if ($form->get('new_draft')->isClicked()) {
+            $versionId = key($form->get('versionIds')->getData());
+            $this->uiVersionService->createDraft($contentId, $versionId);
         }
     }
 }
--- a/src/bundle/Controller/VersionController.php
+++ b/src/bundle/Controller/VersionController.php
@@ -67,7 +67,7 @@ class VersionController extends TabController
         $draftActionsForm->handleRequest($request);
 
         if ($draftActionsForm->isValid()) {
-            $this->deleteVersionsBasedOnFormSubmit($draftActionsForm, $content->id);
+            $this->deleteVersionsBasedOnFormSubmit($draftActionsForm, $content);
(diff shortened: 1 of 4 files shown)


src/lib/Repository/UiVersionService.php
```diff
@@ -30,11 +31,38 @@ public function __construct(ContentService $contentService)
...
+        $versionInfo = $this->loadVersionInfo(
+            $this->loadContentInfo($contentId),
             $versionNo
         );
 
         $this->contentService->deleteVersion($versionInfo);
     }
+
+    /**
+     * Creates a new draft based on the content and verisionNo.
```
glye: versionNo

src/lib/Repository/UiVersionService.php
```diff
@@ -30,11 +31,38 @@ public function __construct(ContentService $contentService)
...
+     *
+     * @param mixed $contentId
+     * @param mixed $versionNo
+     */
+    public function createDraft($contentId, $versionNo)
+    {
+        $contentInfo = $this->loadContentInfo($contentId);
+        $versionInfo = $this->loadVersionInfo(
+            $this->loadContentInfo($contentId),
```
bdunogier: You can use `$contentInfo` from above. And I don't think it's worth wrapping that line then.

src/lib/Repository/UiVersionService.php
bdunogier: Typo: `versionNo`.

src/lib/Repository/UiVersionService.php
```diff
@@ -30,11 +31,38 @@ public function __construct(ContentService $contentService)
...
             $versionNo
         );
 
         $this->contentService->deleteVersion($versionInfo);
     }
+
+    /**
+     * Creates a new draft based on the content and versionNo.
```
bdunogier: `contentId`, actually.

src/bundle/Controller/VersionController.php
```diff
@@ -58,6 +58,7 @@ public function archiveActionsAction(
 
         if ($archiveActionsForm->isValid()) {
             $this->deleteVersionsBasedOnFormSubmit($archiveActionsForm, $content->id);
+            $this->createDraftVersionBasedOnFormSubmit($archiveActionsForm, $content->id);
```
bdunogier: Given that you have a `Content`, is it really worth passing the ID ? It is gonna be loaded back to a `ContentInfo` later on. You might as well pass the `ContentInfo` directly.

mcurran-inviqa (approved):

bdunogier (changes requested):
I'm overall a bit unsure about falling back to id as arguments.

Our Repository's API is on purpose mostly manipulating Value Objects, and I think that we should, when we can, use them instead of scalards in this layer as well.

src/bundle/Form/Versions/ArchivedActions.php
```diff
@@ -15,7 +18,27 @@ public function buildForm(FormBuilderInterface $builder, array $options)
         parent::buildForm($builder, $options);
 
         $builder
-            ->add('delete', SubmitType::class)
+            ->add('delete', SubmitType::class, array(
+                'validation_groups' => false,
```
mcurran-inviqa: use [] syntax for consistency

mcurran-inviqa (approved):

bdunogier (approved):

PR #71, opened 2017-07-12
EZP-27615: If user does not have permission to view versions display message

maxbaldanza:
https://jira.ez.no/browse/EZP-27615

screen shot 2017-07-12 at 15 55 58

### DIFF
--- a/src/bundle/Controller/VersionController.php
+++ b/src/bundle/Controller/VersionController.php
@@ -8,10 +8,10 @@
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
-use EzSystems\HybridPlatformUi\Filter\VersionFilter;
 use eZ\Publish\API\Repository\Values\Content\Content;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
 use EzSystems\HybridPlatformUi\Repository\UiVersionService;
+use EzSystems\HybridPlatformUi\View\Content\Versions\VersionParameterSupplier;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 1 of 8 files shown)


bdunogier:
Should the buttons ("New draft based on select version", "Delete") be hidden when there are no versions displayed ?

maxbaldanza:
Not according to the mock up on the ticket although I think this may be a little weird and have said the same thing on the ticket: https://jira.ez.no/browse/EZP-27615

maxbaldanza:
@bdunogier buttons are now disabled as requested by @SylvainGuittard

SylvainGuittard:
thanks @maxbaldanza

src/bundle/Controller/VersionController.php
```diff
@@ -35,22 +35,19 @@ public function __construct(
...
-        $versions = $this->uiVersionService->loadVersions($contentInfo);
+        $versionParameterSupplier->supply($view);
 
-        $draftVersions = $versionFilter->filterDrafts($versions);
-        $draftActionsForm = $formFactory->createVersionsDraftActionForm($draftVersions);
-
-        $archivedVersions = $versionFilter->filterArchived($versions);
-        $archivedActionsForm = $formFactory->createVersionsArchivedActionForm($archivedVersions);
+        $draftActionsForm = $formFactory->createVersionsDraftActionForm(
+            $view->hasParameter('draftVersions') ? $view->getParameter('draftVersions') : []
```
bdunogier, line 39: I really don't like that way of reusing what the supplier will hopefully set. There is no contract with the interface that certifies any of that.

bdunogier (changes requested):
The way the supplier's data is used in the controller really annoys me. Please try to think of a better approach to this.

src/bundle/Controller/VersionController.php
maxbaldanza: 🤔 not sure there is a better approach without breaking the interface. Did you have something else in mind?
Our views assume that these values are set so isn't it safe for our forms to do the same?

bdunogier (approved):

PR #72, opened 2017-07-12
Do not use variable for translation domain.

maxbaldanza:
Translation extractor cannot read variables therefore have to hardcode
the domain in.
Ran the translation extractor to add in the new translations.


src/bundle/Resources/translations/fieldview.en.xlf.bak
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-07-11T16:39:33Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-07-12T09:14:26Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
```
bdunogier: Could you actually remove the `.bak` files ? I shouldn't have added them.

src/bundle/Resources/translations/fieldview.en.xlf
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-07-11T16:39:33Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
+  <file date="2017-07-12T09:14:26Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
```
bdunogier: I'd remove that file from the diff, as it only has a timestamp update.

src/bundle/Resources/translations/locationview.en.xlf
```diff
@@ -192,69 +192,93 @@
...
+      <trans-unit id="da5ad9f0b2f8de694379f7d3980fc83734e1f48e" resname="locationview.relations.type.link">
+        <source>Link</source>
+        <target state="new">Link</target>
+        <note>key: locationview.relations.type.link</note>
+        <jms:reference-file line="62">src/lib/Templating/Twig/RelationTypeExtension.php</jms:reference-file>
+      </trans-unit>
       <trans-unit id="97d47470f71983df3a61744fb5c0b2c5d51774d5" resname="locationview.translations.languagecode">
         <source>Language code</source>
-        <target state="new">Language code</target>
+        <target>Language code</target>
```
bdunogier: Could you cleanup the change so that only the added strings (block above) are in the diff ?

bdunogier (changes requested):

bdunogier (approved):

PR #73, opened 2017-07-12
EZP-27538 delete translations

mcurran-inviqa:
- Adds translation action form
- Moves translation tab view into TranslationController
- Handles deletion of translations via action form

image

https://jira.ez.no/browse/EZP-27538


maxbaldanza:
I've spoken to @andrerom previously about updating the kernel version and because `PlatformUIBundle` requires V6 we can't update to V7: https://github.com/ezsystems/PlatformUIBundle/blob/master/composer.json#L14
This dependancy should be updated to `^6.11@dev|^7.0@dev` to allow us to use V7.

maxbaldanza (approved):

bdunogier (approved):

PR #75, opened 2017-07-13
EZP-27581: Add content type name to the content view

maxbaldanza:
https://jira.ez.no/browse/EZP-27581

screen shot 2017-07-13 at 10 46 58

### DIFF
--- a/spec/View/Content/ContentType/ContentTypeParameterSupplierSpec.php
+++ b/spec/View/Content/ContentType/ContentTypeParameterSupplierSpec.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace spec\EzSystems\HybridPlatformUi\View\Content\ContentType;
+
+use eZ\Publish\API\Repository\ContentTypeService;
+use eZ\Publish\API\Repository\Values\Content\ContentInfo;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+use eZ\Publish\Core\MVC\Symfony\View\ContentView;
+use eZ\Publish\Core\Repository\Values\Content\Content;
+use eZ\Publish\Core\Repository\Values\Content\VersionInfo;
+use EzSystems\HybridPlatformUi\View\Content\ParameterSupplier;
+use PhpSpec\ObjectBehavior;
+
+class ContentTypeParameterSupplierSpec extends ObjectBehavior
+{
+    function let(ContentTypeService $contentTypeService)
+    {
+        $this->beConstructedWith($contentTypeService);
+    }
+
+    function it_is_initializable()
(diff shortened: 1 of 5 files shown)


src/lib/View/Content/ContentType/ContentTypeParameterSupplier.php
```diff
@@ -0,0 +1,38 @@
...
+    {
+        $this->contentTypeService = $contentTypeService;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function supply(ContentView $contentView)
+    {
+        $contentType = $this->contentTypeService->loadContentType(
```
maxbaldanza, line 32: Should we be catching the exception that may be thrown here? If so what should the outcome be

src/bundle/Resources/views/locationview.html.twig
```diff
@@ -6,6 +6,7 @@
         {% block contentview_details %}
         <header class="ez-page-header">
             <h1>{{ ez_content_name(content) }}</h1>
+            <h2>{{ contentType.name }}</h2>
```
dpobel: I think this should not be in a `h2`, I would propose the following markup:

```html
    <h1>
        {{ ez_content_name(content) }}
        <br>
        <span class="ez-content-type-name">{{ contentType.name }}</span>
    </h1>
</header>
```

src/lib/View/Content/ContentType/ContentTypeParameterSupplier.php
dpobel: well not really I think. It seems like it can only throw a NotFoundException and if it's the case that's a very bad sign for the database consistency ;-) and we can not do anything about that so I guess it's ok like this

bdunogier (approved):
Looks good to me, with the markup changes requested by @dpobel.

spec/View/Content/ContentType/ContentTypeParameterSupplierSpec.php
```diff
@@ -0,0 +1,44 @@
...
+use PhpSpec\ObjectBehavior;
+
+class ContentTypeParameterSupplierSpec extends ObjectBehavior
+{
+    function let(ContentTypeService $contentTypeService)
+    {
+        $this->beConstructedWith($contentTypeService);
+    }
+
+    function it_is_initializable()
```
mcurran-inviqa: this should be renamed to it_is_a_supplier it's actually an interface check

mcurran-inviqa (approved):

dpobel (approved):

bdunogier (approved):

PR #77, opened 2017-07-13
EZP-27579 add breadcrumbs to content view

maxbaldanza:
- [X] Move loading locations to ParameterSupplier so that it can be used for loading the breadcrumbs.
- [X] Add breadcrumbs to content view

https://jira.ez.no/browse/EZP-27579

screen shot 2017-07-14 at 10 01 13

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -15,6 +15,7 @@ use eZ\Publish\API\Repository\Values\Content\Location;
 use eZ\Publish\Core\MVC\Symfony\View\ContentView;
 use EzSystems\HybridPlatformUi\Form\UiFormFactory;
 use EzSystems\HybridPlatformUi\Repository\UiLocationService;
+use EzSystems\HybridPlatformUi\View\Content\Locations\LocationParameterSupplier;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\RouterInterface;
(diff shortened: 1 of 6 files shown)


src/bundle/Resources/views/locationview.html.twig
```diff
@@ -4,6 +4,12 @@
     <section class="ez-content-view-column ez-tabs">
     {% block contentview %}
         {% block contentview_details %}
+        {% for location in locations %}
+            {% if location.id == locationId %}
+                {{ include('@EzSystemsHybridPlatformUi/fields/location/path.html.twig', {'locations': location.pathLocations}) }}
```
dpobel: put that in the `header` below and before the `h1`

dpobel (changes requested):

src/bundle/Controller/LocationController.php
```diff
@@ -43,17 +44,20 @@ public function __construct(
...
         $versionInfo = $view->getContent()->getVersionInfo();
         $contentInfo = $versionInfo->getContentInfo();
 
         if ($contentInfo->published) {
-            $locations = $this->uiLocationService->loadLocations($contentInfo);
-            $actionsForm = $this->formFactory->createLocationsActionForm($locations);
+            $locationParameterSupplier->supply($view);
+
+            $actionsForm = $this->formFactory->createLocationsActionForm(
+                $view->hasParameter('locations') ? $view->getParameter('locations') : []
```
bdunogier, line 52: I understand why you would reuse that value, but it really feels like a sever break of responsibilities.

bdunogier (approved):
Not a huge fan, but it gets the job done. The suppliers look a bit superfluous in this implementation.

mcurran-inviqa (approved):

bdunogier (approved):

PR #78, opened 2017-07-17
EZP-27448 url aliases tab

mcurran-inviqa:
- adds url controller
- adds paramter suppliers for urls
- adds view only url aliases tab

image

https://jira.ez.no/browse/EZP-27448

### DIFF
--- a/src/bundle/Resources/views/content/tabs/urls.html.twig
+++ b/src/bundle/Resources/views/content/tabs/urls.html.twig
@@ -18,9 +18,7 @@
         {% for customUrl in customUrls %}
             <tr>
                 <td></td>
-                <td>
-                    <a href="{{ customUrl.path }}">{{ customUrl.path }}</a>
-                </td>
+                <td>{{ customUrl.path }}</td>
                 <td>
                     {% for languageCode in customUrl.languageCodes %}{{ languageCode }}<br>{% endfor %}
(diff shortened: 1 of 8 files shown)


maxbaldanza (approved):

src/bundle/Resources/views/content/tabs/urls.html.twig
```diff
@@ -0,0 +1,75 @@
...
+            <th>{{ 'locationview.urls.language'|trans|desc('Language') }}</th>
+            <th>{{ 'locationview.urls.type'|trans|desc('Type') }}</th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for customUrl in customUrls %}
+            <tr>
+                <td></td>
+                <td>
+                    <a href="{{ customUrl.path }}">{{ customUrl.path }}</a>
```
dpobel: according to description in https://jira.ez.no/browse/EZP-27448, this is not supposed to be a link (which leads to a different siteaccess btw)

dpobel (changes requested):

dpobel (approved):

bdunogier:
When the same URL alias is shared by multiple languages, the lines are supposed to be merged.

bdunogier (approved):

PR #79, opened 2017-07-17
EZP-27506 toggle locations

mcurran-inviqa:
- Adds location visibility form
- Adds visibility fields to actions form for populating location visibility form
- Adds handler for visibility toggle

image

Additonally some js work is needed to populate and submit the hidden notification status form and handle notification responses.

https://jira.ez.no/browse/EZP-27506

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -52,10 +52,12 @@ class LocationController extends TabController
                 $view->hasParameter('locations') ? $view->getParameter('locations') : []
             );
             $swapLocationsForm = $this->formFactory->createLocationsContentSwapForm();
+            $visibilityForm = $this->formFactory->createLocationVisibilityForm();
 
             $view->addParameters([
                 'actionsForm' => $actionsForm->createView(),
                 'swapLocationsForm' => $swapLocationsForm->createView(),
+                'visibilityForm' => $visibilityForm->createView(),
             ]);
         }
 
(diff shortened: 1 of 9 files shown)


src/lib/Notification/NotificationPoolAwareTrait.php
```diff
@@ -82,4 +83,16 @@ protected function addErrorNotification(string $message, array $params = [], $do
 
         $this->notificationPool->add($notification);
     }
+
+    /**
+     * Send last notification message as a response.
+     */
+    protected function notificationResponse()
```
maxbaldanza, line 90: This may be better as a public method as it's being used from the controller?

src/bundle/Controller/LocationController.php
```diff
@@ -122,6 +124,41 @@ public function updateDefaultSortOrderAction(
...
+        Request $request,
+        LocationService $locationService,
+        UiFormFactory $formFactory
+    ) {
+        try {
+            $visibilityForm = $formFactory->createLocationVisibilityForm();
+            $visibilityForm->handleRequest($request);
+
+            if (!$visibilityForm->isValid()) {
+                throw new \Exception(var_export($visibilityForm->getErrors(), true));
```
maxbaldanza, line 136: Should this be changed to use `addErrorNotification`?

maxbaldanza (changes requested):

src/bundle/Controller/LocationController.php
mcurran-inviqa: it's caught and added as an error notification.

src/lib/Notification/NotificationPoolAwareTrait.php
mcurran-inviqa: it's a trait the controller has no issue with it being protected, it should not be used outside of the controller

src/bundle/Controller/LocationController.php
maxbaldanza: 👍 must have missed that

bdunogier:
Rebased on top of master.

@dpobel this needs the JS features that were mentioned when this feature was designed (posting the form when the checkbox is toggled).

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -38,7 +38,7 @@
                     {% endif %}
                 {% endif %}
             </td>
-            <td></td>
+            <td>{{ form_widget(actionsForm.locationVisibility[location.id]) }}</td>
```
dpobel, line 41: the field is completely useless in the `Location/Action` form. Since we only need a checkbox (and actually most likely more than that to have the widget slider to switch visibility) to trigger a submit of the `Location/Visibility` form, I  think we should remove the `locationVisibility` field from `Location/Action` and here directly generate a (disabled) checkbox input reflecting the visibility which we'll use to submit the other form. In addition, according to https://jira.ez.no/browse/EZP-27506, we should also display there if the Location is hidden because of one of its ascendant, so in the end we should have something like:

```twig
<input type="checkbox" name="visibility_{{ location.id }}" value="1" disabled
    {% if location.hidden %}checked{% endif %}>
{% if location.invisible %}
    <p>Hidden by superior</p>
{% endif %}
```

bdunogier: You are right when you say that the visibility form item has nothing to do with `Location/Action`. But it also bothers me a bit that it has nothing to do with `Location/Visibility` :)

I'm still unsure wether client side form items such as this one should be known to the application layer or not. It would make sense if the location id form widgets were provided by the `Location\Visibility` form object:

```twig
{{ form_widget(visibilityForm.locationVisibility[location.id]) }}
```

The Form would know that it needs to generate a checkbox, or whatever the frontend needs, and the template wouldn't have to know about it.

An alternative is to consider that the frontend code's specific requirements are first and foremost configured in templates. But the fact is that for now, we have a good part of it in the FormTypes from the package.

PR #17, opened 2017-06-15
EZP-27109: update the Symfony Web Debug Toolbar during navigation

bdunogier:
> [EZP-27109](http://jira.ez.no/browse/EZP-27109)
> Replaces https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/15
> Requires https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/17

Listens to the `ez:app:updated` event, and reloads the Symfony Web Debug Toolbar after update requests.

PR #85, opened 2017-07-18
 EZP-27573: Confirm mechanism before removing a Location or a version

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27573
Requires: https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/31
Requires: ezsystems/hybrid-platform-ui-assets version 0.4

# Description

This PR and https://github.com/ezsystems/PlatformUIBundle/pull/885 adds a confirm dialog when using a *remove something* button in Hybrid UI.

Screencast:

[](http://www.youtube.com/watch?v=SoeAIbOUsrA "Click to play on Youtube.com")

# Test

manual tests


StephaneDiot (approved):

src/bundle/Resources/public/css/modules/modal.css
```diff
@@ -0,0 +1,31 @@
...
+    background: var(--ez-color-ground-modal);
+
+    box-shadow: 0 3px 20px var(--ez-color-shadow-modal);
+}
+
+.ez-modal header, .ez-modal section, .ez-modal footer {
+    padding: 1em;
+}
+
+.ez-modal header {
```
SylvainGuittard: The `confirm-dialog` component does not have the `header` part.
Do we need a follow-up issue for this?

bdunogier:
Screencast: is it me, or the notification says that the section was _created_ when it gets deleted ?

Besides that question, I'm reviewing it.
Trying to think of simplifications for the twig part.

yannickroger:
It is not you @bdunogier :) https://youtu.be/SoeAIbOUsrA?t=36s
There is an issue with the notification.

src/bundle/Resources/public/css/modules/modal.css
yannickroger: The header part was not in the example you provided in JIRA, so I guess it is a left over from a previous version of the code. Is that it @dpobel ?

yannickroger (approved):
Approved once the header thing is sorted out.

src/bundle/Resources/public/css/modules/modal.css
dpobel: @SylvainGuittard @yannickroger it's not a left over, it's made on purpose because `.ez-modal` class will be used for others modals than the confirm dialog in the app and [the UI Guidelines specifies that a modal can have a header](https://ezsystems.github.io/ui-guidelines/#modal3)

dpobel:
@bdunogier @yannickroger on the notification, it's you :p Actually, the notification appears right after creating a Section but it's hidden by the SF toolbar (if you look carefully at https://www.youtube.com/watch?v=SoeAIbOUsrA&feature=youtu.be&t=32s, you can see a very small green part of it), then when you remove a Section, another notification is pushed below the previous one, that's why you see it at that moment.

The real issue here is that the SF toolbar is buggy as soon as you manually reload it with its `load` method (Done in https://github.com/ezsystems/hybrid-platform-ui/pull/17), it can not be minimized again, its display is buggy and  if it was minimized, it becomes maximized. (~~I'm reporting an issue for that~~ https://github.com/symfony/symfony/issues/23609)

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -1,12 +1,31 @@
 {% trans_default_domain "locationview" %}
 {{ form_start(actionsForm, {'action': path('ez_hybrid_platform_ui_location_actions', {'contentId': content.id, 'redirectLocationId': location.id})}) }}
 
+<dialog id="confirm-location-removal-{{ location.id }}" class="ez-modal">
+    <section>
+        <p>{{ 'confirm.remove.location'|trans()|desc('Are you sure you want to remove selected locations?') }}</p>
+    </section>
+    <footer>
+        <button type="button" class="ez-button ez-button-secondary ez-js-close-modal">
```
inakijv: For this kind of modal (alerting the possibility of removing content) we suggested that `Cancel` buttons are styled just as `.ez-button` and not `.ez-button-secondary`

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -1,12 +1,31 @@
...
+            {{ 'confirmbox.cancel'|trans({}, 'confirmbox')|desc('Cancel') }}
+        </button>
+        {{
+            form_widget(actionsForm.delete, {
+                'label': 'confirmbox.confirm'|trans({}, 'confirmbox')|desc('Confirm'),
```
inakijv: For this kind of modal instead of a `Confirm` button we have suggested that the text of that button should be `Delete`

src/bundle/Resources/views/content/tabs/locations.html.twig
```diff
@@ -1,12 +1,31 @@
...
+                'attr': {'class': 'ez-button ez-button-primary ez-js-close-modal'}
```
inakijv: If the text of the button is now `Delete`, then the style of the button should change from `.ez-button-primary` to `.ez-button-negative`

inakijv:
Regarding the modals' design and animation, the UI Guidelines include a few suggestions we would like to discuss/share with you (but maybe you are aware of them :-) ):

_Design_.
In terms of extensibility purposes (I am not sure if we are including this factor here) shouldn't we define sort of a max-width (we suggested 80%) and a max-height (suggested 70vh) for them, should we? Given that we have at least 3 different types of modals (default modal, modal with header and modal with header and buttons) we need to be sure that this is possible to apply for all cases. 

Besides of this, we have suggested a standard `padding:  1em 3.33em;` for the content within the modal, that also applies for the buttons despite of being right aligned.

_Animation_.
Another proposal is to define the same animation for all of them (the UI Guidelines suggests which one it is), as well as place them at the same distance from the top of the browser window. The suggested distance is `top: 75px;`

src/bundle/Resources/views/content/tabs/locations.html.twig
dpobel: > For this kind of modal (alerting the possibility of removing content) we suggested that `Cancel` buttons are styled just as `.ez-button` and not `.ez-button-secondary`

ok I'll do it but @inakijv you suggested that where ?

For the record, when this was implemented the only spec I had was this comment https://jira.ez.no/browse/EZP-27573#comment-213523

src/bundle/Resources/views/content/tabs/locations.html.twig
dpobel: same as above but ok

src/bundle/Resources/views/content/tabs/locations.html.twig
dpobel: and same as well

dpobel:
@inakijv for the *Design.* part ok. For *Animation.*, I don't understand. The very same ~~animation~~ transition is implemented in https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/31/files#diff-e8765e08d4a05c6fbdcbf1bdae00f8ecR3 look at the screencast. For the position, I'll do it as well (even if I personally prefer to have it centered on the screen, that way, the confirm button is in average closer to the user's pointer)

src/bundle/Resources/views/content/tabs/locations.html.twig
inakijv: You did it right @dpobel, but in that part the specification is not correct. And yes, I did not ask for that previously, but implicitly the UI Guidelines include this example as one of the three types of modals. 

Sorry for not asking you before about this, but I was on vacation at that time :-(

inakijv:
@dpobel thanks for your feedback on the animation, I am glad that you approved and added the latest suggestion :-) (I am referring to [this one](https://github.com/ezsystems/hybrid-platform-ui/pull/85#issuecomment-317025267)) My comment wasn't addressed to the current modal you are implementing, but for the ones to come. 

With the implementation of the UI Guidelines we aim to create some patterns, one of them is for modal windows. Given that scanning comes before clicking, by placing them at the same distance, no matter what the content is, users will have the certainty of what is going to come up next and where to look up on the screen.

PR #87, opened 2017-07-24
EZP-27657: Use the UDW to add Locations

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27657
Requires: https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/35 and https://github.com/ezsystems/PlatformUIBundle/pull/887

# Description

* [x] Allow several Locations to be added at once
* [x] Use the convention to run the UDW

Most of the work is done in https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/35 where the markup convention to run the UDW is defined and recognized.

Screencast:

[](http://www.youtube.com/watch?v=cr7fnJU8G3k "Click to play on Youtube.com")

# Test

manual test

### DIFF
--- a/src/bundle/Controller/LocationController.php
+++ b/src/bundle/Controller/LocationController.php
@@ -78,7 +78,7 @@ class LocationController extends TabController
                 return $this->resetToMainLocation($content->id);
             }
 
-            $this->addLocationBasedOnFormSubmit($actionsForm, $content);
+            $this->addLocationsBasedOnFormSubmit($actionsForm, $content);
         }
 
         return $this->reloadTab('locations', $content->id, $redirectLocationId);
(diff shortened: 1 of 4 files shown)


src/bundle/Controller/LocationController.php
```diff
@@ -144,12 +144,16 @@ private function deleteLocationsBasedOnFormSubmit(FormInterface $form, $redirect
...
     }
 
-    private function addLocationBasedOnFormSubmit(FormInterface $form, Content $content)
+    private function addLocationsBasedOnFormSubmit(FormInterface $form, Content $content)
     {
         if ($form->get('add')->isClicked()) {
-            $this->uiLocationService->addLocation(
-                $content->contentInfo,
-                $form->get('parentLocationId')->getData()
+            $locationService = $this->uiLocationService;
```
bdunogier: You don't need to do that, you can use `$this->uiLocationService` from the callback.

bdunogier:
https://github.com/ezsystems/hybrid-platform-ui/pull/86 is a good addition to that :)

src/bundle/Controller/LocationController.php
```diff
@@ -144,12 +144,16 @@ private function deleteLocationsBasedOnFormSubmit(FormInterface $form, $redirect
...
+            $contentInfo = $content->contentInfo;
+            array_map(
+                function ($locationId) use ($locationService, $contentInfo) {
+                    $locationService->addLocation($contentInfo, (int) $locationId);
+                },
+                explode(',', $form->get('parentLocationId')->getData())
```
adamwojs: I'm not involved directly into hybryd ui development, but I think that this kind of logic should be moved from controller to a Form Definition level as DataTransformer. Skip this comment if there are other assumptions in project.

dpobel:
~~(investigating Travis CI failure)~~ Caused by 2 issues:

1. Hybrid Platform UI depends on PlatformUIBundle which depends on Hybrid Platform UI. "Fixed" in e9e9f7d
1. We are experiencing the bug https://github.com/travis-ci/travis-ci/issues/8113, workaround in 269145f

yannickroger (approved):

bdunogier (approved):
It would benefit from some Symfony Forms love to ease the usage of it, but still good.

PR #88, opened 2017-07-24
Ran translation update

yannickroger:
I ran the command for a PR I am working. Since there was a backlog of diff, I didn't want it there.

### DIFF
--- a/src/bundle/Resources/translations/locationview.en.xlf
+++ b/src/bundle/Resources/translations/locationview.en.xlf
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" xmlns:jms="urn:jms:translation" version="1.2">
-  <file date="2017-07-11T16:39:33Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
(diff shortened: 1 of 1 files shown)


src/bundle/Resources/translations/locationview.en.xlf
```diff
@@ -1,14 +1,14 @@
...
+  <file date="2017-07-24T15:35:28Z" source-language="en" target-language="en" datatype="plaintext" original="not.available">
     <header>
       <tool tool-id="JMSTranslationBundle" tool-name="JMSTranslationBundle" tool-version="1.1.0-DEV"/>
       <note>The source node in most cases contains the sample message as written by the developer. If it looks like a dot-delimitted string such as "form.label.firstname", then the developer has not provided a default message.</note>
     </header>
     <body>
       <trans-unit id="e0be1422e1af400879a6716220eabc9ad4fca29e" resname="locationview.details.ascending">
         <source>Ascending</source>
-        <target state="new">Ascending</target>
+        <target>Ascending</target>
```
bdunogier, line 11: These shouldn't go in. Not sure if we need to post-process the file or prevent this change (you could check the options).

An alternative is to `git add -p` the changes, and only add the real changes.

bdunogier (changes requested):
The removal of `state="new"` should be prevented if possible.

yannickroger:
So we'll run this script only before releasing so "new" state will be kept.

PR #89, opened 2017-07-24
EZP-27671: Send a content item to the trash

yannickroger:
Link: https://jira.ez.no/browse/EZP-27671

## Description
Adds a "Send to Trash" button to delete a location.

## Tests
Manual tests

## TODO
- [x] Send to trash button
- [x] Confirm box
- [x] Confirm message with childcount
- [x] Disable button when user doesn't have the permission
- [x] Cleanup and rebase

### DIFF
--- a/bin/extract-translations.sh
+++ b/bin/extract-translations.sh
@@ -11,6 +11,6 @@ echo '# Extracting translations from vendor/ezsystems/hybrid-platform-ui';
 cd vendor/ezsystems/hybrid-platform-ui;
 
 echo '# Clean file references';
-sed -i .bak "s|>.*/hybrid-platform-ui/|>|g" ./src/bundle/Resources/translations/*.xlf
+sed -i.bak "s|>.*/hybrid-platform-ui/|>|g" ./src/bundle/Resources/translations/*.xlf
 
 echo 'Translation extraction done';
--- a/spec/Repository/UiLocationServiceSpec.php
+++ b/spec/Repository/UiLocationServiceSpec.php
@@ -8,6 +8,7 @@ namespace spec\EzSystems\HybridPlatformUi\Repository;
 
 use eZ\Publish\API\Repository\ContentTypeService;
 use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\TrashService;
 use eZ\Publish\API\Repository\Values\Content\ContentInfo;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentException;
 use eZ\Publish\API\Repository\Values\Content\LocationCreateStruct;
@@ -23,11 +24,18 @@ class UiLocationServiceSpec extends ObjectBehavior
 {
     function let(
         LocationService $locationService,
+        TrashService $trashService,
         PathService $pathService,
         UiPermissionResolver $permissionResolver,
         ContentTypeService $contentTypeService
     ) {
-        $this->beConstructedWith($locationService, $pathService, $permissionResolver, $contentTypeService);
+        $this->beConstructedWith(
+            $locationService,
+            $trashService,
+            $pathService,
+            $permissionResolver,
+            $contentTypeService
+        );
     }
 
     function it_loads_ui_locations_with_a_child_count(LocationService $locationService)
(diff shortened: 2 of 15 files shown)


bin/extract-translations.sh
```diff
@@ -11,6 +11,6 @@ echo '# Extracting translations from vendor/ezsystems/hybrid-platform-ui';
 cd vendor/ezsystems/hybrid-platform-ui;
 
 echo '# Clean file references';
-sed -i .bak "s|>.*/hybrid-platform-ui/|>|g" ./src/bundle/Resources/translations/*.xlf
+sed -i "s|>.*/hybrid-platform-ui/|>|g" ./src/bundle/Resources/translations/*.xlf
```
bdunogier: Well, without this, it fails on mac... :)

There's probably a better solution, but I don't remember it.

src/bundle/Form/Locations/LocationTrash.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        parent::buildForm($builder, $options);
+
+        $builder
+            ->add('trash', SubmitType::class);
+    }
+
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        parent::configureOptions($resolver);
```
bdunogier: Not needed, the parent is empty, and will remain so by definition.

bdunogier (commented):
Looks good so far.

spec/Repository/UiLocationServiceSpec.php
```diff
@@ -108,6 +108,68 @@ function it_deletes_locations(LocationService $locationService)
         $this->deleteLocations([$deleteLocationId]);
     }
 
+    function it_trashes_a_single_location_and_returns_the_parent(LocationService $locationService)
```
bdunogier: I'm not really set on the method returning the parent. Isn't it the controller's responsibility ?

No strong conviction, but maybe you had considered this.

src/bundle/Form/Locations/LocationTrash.php
```diff
@@ -0,0 +1,29 @@
...
+use Symfony\Component\Form\AbstractType;
+use Symfony\Component\Form\Extension\Core\Type\SubmitType;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class LocationTrash extends AbstractType
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        parent::buildForm($builder, $options);
```
bdunogier: You don't need that.

bdunogier (approved):
I'm fine with it.

About permissions & forms, I have done a spike on that last week: https://github.com/ezsystems/hybrid-platform-ui/commit/7e3f3eae81de4d59c56360c0ab72270dec5f87fb. It moves permissions checking to the form itself.

You may want to have a look at it. It's not a must for this PR, but it's strongly conneced to it.

spec/Repository/UiLocationServiceSpec.php
yannickroger: @bdunogier I did consider it. It was in the controller when I first wrote the code. But I ended up injecting the location service and I thought it could be nice to have to have smaller controller. The service makes this code reusable and testable.
Your comment made me realize the method was poorly named, I improved it, tell me what you think :)

src/bundle/Resources/public/css/modules/button.css
```diff
@@ -81,3 +81,16 @@
...
+    white-space: normal;
+    width: 100%;
+    text-align: center;
+
+    background: transparent;
+    border-color: transparent;
+}
+
+.ez-button-action:hover:not([disabled]) {
+    background: rgba(0, 0, 0, 0.2);;
```
SylvainGuittard: extra `;`

src/bundle/Resources/public/css/modules/button.css
```diff
@@ -81,3 +81,16 @@
     background: var(--ez-color-secondary);
     color: var(--ez-color-text-base-inverted);
 }
+
+.ez-button-action {
+    white-space: normal;
+    width: 100%;
+    text-align: center;
+
```
SylvainGuittard: Why an extra line?

src/lib/Components/Browse.php
```diff
@@ -30,7 +30,7 @@ public function renderToString()
         $selected = $this->getLocationId();
         // should be rendered with a twig template
         return '<ez-browse
-            class="ez-button"
+            class="ez-button  ez-button-action"
```
SylvainGuittard, line 33: This button is in the discovery bar (left menu). I am not sure it should be in this PR.
But we can keep this change if you want :)

src/lib/Components/Search.php
```diff
@@ -6,7 +6,7 @@ class Search implements Component
 {
     public function renderToString()
     {
-        return '<button class="ez-button" disabled>Search</button>';
+        return '<button class="ez-button ez-button-action" disabled>Search</button>';
```
SylvainGuittard, line 9: same as https://github.com/ezsystems/hybrid-platform-ui/pull/89/files#r130595775

src/lib/Components/Trash.php
```diff
@@ -6,7 +6,7 @@ class Trash implements Component
 {
     public function renderToString()
     {
-        return '<button class="ez-button" disabled>
+        return '<button class="ez-button  ez-button-action" disabled>
```
SylvainGuittard, line 9: same as https://github.com/ezsystems/hybrid-platform-ui/pull/89/files#r130595775

src/lib/Components/Browse.php
yannickroger: It fixes the bug there :) I could put it in a dedicated commit, but since I am doing CSS on the buttons it's not a big deal. I would have done it if it hard to be backported.

SylvainGuittard: ok :)
remove the extra space ;)

src/lib/Components/Trash.php
SylvainGuittard: CS: remove the extra space

bin/extract-translations.sh
dpobel: it's because there's a small difference in the `-i` option between BSD sed (the one on Mac) and GNU sed (usually the one on Linux). To have something which works well in both case, this line should be written:

```shell
sed -i.bak "s|>.*/hybrid-platform-ui/|>|g" ./src/bundle/Resources/translations/*.xlf
```

(no space between `-i` and `.bak`)

dpobel (approved):
+1

PR #91, opened 2017-07-25
EZP-27701: enabled the Versions tab 'Edit' button for drafts

bdunogier:
> [EZP-27701](http://jira.ez.no/browse/EZP-27701)

Title says it all.

Doesn't include checking for permissions. Adding that will require more changes.

### DIFF
--- a/src/bundle/Resources/views/content/tabs/versions/table.html.twig
+++ b/src/bundle/Resources/views/content/tabs/versions/table.html.twig
@@ -41,7 +41,7 @@
             <td>
                 {{ version.modificationDate|localizeddate('medium', 'short', app.request.locale) }}
             </td>
-            {% if isDraft is defined %}
+            {% if isDraft is defined and version.canEdit %}
                 <td>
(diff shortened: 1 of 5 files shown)


bdunogier:
Permissions checking added. The button won't be shown if the user isn't allowed to edit.

andrerom (approved):
Passing, @dpobel who should review this?

src/bundle/Resources/views/content/tabs/versions/table.html.twig
```diff
@@ -41,9 +41,13 @@
             <td>
                 {{ version.modificationDate|localizeddate('medium', 'short', app.request.locale) }}
             </td>
-            {% if isDraft is defined %}
+            {% if isDraft is defined and version.canEdit %}
```
dpobel, line 44: the `version.canEdit` check should be separated to affect only the generation (or not) of the link, otherwise if the user can not edit a version, the cell will be missing in the table.

dpobel (changes requested):

PR #92, opened 2017-07-26
Add priority to navigation tags so we are able to set navigation display order

Nattfarinn:
Right now `NavigationHub` links (and zones) are in order of bundle loading, and because of most of our bundles use `PrependInterface` it's even more difficult to position new navigation options as we like.

This implementation adds `priority` (lower first) handling to `ezplatform.ui.link` and `ezplatform.ui.zone` compiler pass.

### DIFF
--- a/docs/howto/add_navigation_hub_items.md
+++ b/docs/howto/add_navigation_hub_items.md
@@ -114,3 +114,18 @@ A setter method `setRoutePrefix` can be used to define a route prefix used to ma
 If given, a Request's route with the same prefix will match (example: `admin_sectionedit` matches the
 `admin_section` prefix).
 
+## Position
+Order of both, links and zones, can be set. It's implemented in a *Symfony way* as a `priority` parameter of service definition tag:
+
+```yaml
+services:
+    ezsystems.platformui.navigationhub.link.admin.dashboard:
+        # ...
+        tags:
+            - name: ezplatform.ui.link
+              priority: 100
+```
+
+Lower `priority` values will make links appear first, those with the same `priority` value will appear in an order of definition, higher values obviously later.
+Good practice is to make default values multiplier of `10` or `100`, so end-user will be able to inject own links (or zones) between default ones (ie. by setting `priority` to `50`).
+Default value of omitted `priority` parameter is `0`. Negative values are accepted as well.
--- a/src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
+++ b/src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
@@ -34,7 +34,7 @@ class NavigationHubPass implements CompilerPassInterface
         }
 
         $sorted = [];
-        krsort($unsorted);
+        ksort($unsorted);
         foreach ($unsorted as $items) {
             $sorted = array_merge($sorted, $items);
         }
--- a/src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
+++ b/src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
@@ -24,14 +24,23 @@ class NavigationHubPass implements CompilerPassInterface
 
     private function processNavigationHubTag(ContainerBuilder $container, $tag, $index)
     {
-        $items = [];
+        $unsorted = [];
         $services = $container->findTaggedServiceIds($tag);
-        foreach (array_keys($services) as $serviceId) {
-            $items[] = new Reference($serviceId);
+        foreach ($services as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                $priority = isset($tag['priority']) ? $tag['priority'] : 0;
+            }
+            $unsorted[$priority][] = new Reference($serviceId);
+        }
+
+        $sorted = [];
+        krsort($unsorted);
+        foreach ($unsorted as $items) {
+            $sorted = array_merge($sorted, $items);
         }
 
         $container
             ->findDefinition(NavigationHub::class)
-            ->replaceArgument($index, $items);
+            ->replaceArgument($index, $sorted);
     }
 }
--- a/src/bundle/Resources/config/navigationhub.yml
+++ b/src/bundle/Resources/config/navigationhub.yml
@@ -15,7 +15,8 @@ services:
             - "Content"
             - "content"
         tags:
-            - {name: ezplatform.ui.zone}
+            - name: ezplatform.ui.zone
+              priority: 100
 
     ezsystems.platformui.navigationhub.zones.page:
         class: "%ezsystems.platformui.navigationhub.zone.class%"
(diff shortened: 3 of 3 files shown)


src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
```diff
@@ -24,14 +24,23 @@ public function process(ContainerBuilder $container)
...
     private function processNavigationHubTag(ContainerBuilder $container, $tag, $index)
     {
-        $items = [];
+        $unsorted = [];
         $services = $container->findTaggedServiceIds($tag);
-        foreach (array_keys($services) as $serviceId) {
-            $items[] = new Reference($serviceId);
+        foreach ($services as $serviceId => $tags) {
+            foreach ($tags as $tag) {
+                $priority = isset($tag['priority']) ? $tag['priority'] : 0;
```
dpobel, line 31: instead of having a default priority, I think those services (including ours) should always have one to make the whole a process a bit more predictable.

dpobel (changes requested):
In addition to the inline comment, you should also update the corresponding doc https://github.com/ezsystems/hybrid-platform-ui/blob/master/docs/howto/add_navigation_hub_items.md

bdunogier:
I'm mostly wondering if "index" or "order" wouldn't make more sense than "priority" for these items. Descending order sounds odd for ordering.

src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
bdunogier: I think I can agree with that.

src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
```diff
@@ -24,14 +24,23 @@ public function process(ContainerBuilder $container)
...
+            }
+            $unsorted[$priority][] = new Reference($serviceId);
+        }
+
+        $sorted = [];
```
bdunogier, line 36: Nitpick: a private method that you call directly in `replaceArgument` would be more readable.

bdunogier (approved):

bdunogier:
Could you also open a JIRA story and update the PR / commit to refer to it ?

src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
Nattfarinn: But `replaceArgument` is a method of Symfony definition. It alters Dependency Injection definition with new argument value.

bdunogier: > But replaceArgument is a method of Symfony definition.

I know, no worries :)
I meant:

```
$container
  ->findDefinition(NavigationHub::class)
  ->replaceArgument(
    $index,
    $this->mapPrioritizedItemsList($container, $services)
  );
```

But again, it's more for readability of the code. Even though what it does isn't that hard to grasp.

andrerom:
> I'm mostly wondering if "index" or "order" wouldn't make more sense than "priority" for these items. Descending order sounds odd for ordering.

isn't `priority` kind of a symfony convention? And one we even use ourselves some places?

bdunogier:
in an event context, absolutely andré. For UI elements i was wondering what made the most sense.

andrerom:
> in an event context, absolutely andré. For UI elements i was wondering what made the most sense.

clear. well maybe @emodric or @lolautruche knows of other ways that can serve as example here.

emodric:
`priority` sounds okay to me. Built in services should have a priority defined with priorities set at 100, 200, 300 and so on, to allow injecting menu items in between them.

lolautruche:
I agree with @emodric

src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
Nattfarinn: ~~Ah, OK. I thought about it yet, I didn't like the idea of making private method call to another private method when it's called only once. But I guess you're right. :)~~

I made a few attempts to refactor it, but each made code (while somewhat readable) a bit... spaghetti.

src/bundle/DependencyInjection/Compiler/NavigationHubPass.php
Nattfarinn: Having default value reduces effort required to make extensions, while still gives an option to personalise it as you like. I would opt for a compromise here and leave default value, but still explicitly declare it for our services to make extending predictable.

Nattfarinn:
@dpobel True. I forgot about the docs (how-to updated).
@bdunogier I thought about different name, but `priority` is... hmm... neutral. Names like `position`, ~~`order`,~~ `index` suggest **exact** location, while all it does is sorting by this value. 

Edit: I guess `order` would be best alternative, but `priority` is *kind of* standard. It's up to you guys.

docs/howto/add_navigation_hub_items.md
```diff
@@ -114,3 +114,18 @@ A setter method `setRoutePrefix` can be used to define a route prefix used to ma
 If given, a Request's route with the same prefix will match (example: `admin_sectionedit` matches the
 `admin_section` prefix).
 
+## Position
+Order of both, links and zones, can be set. It's implemented in a *Symfony way* as a `priority` parameter of service definition tag:
```
dpobel, line 118: could you please wrap lines like in previous paragraphs ?

docs/howto/add_navigation_hub_items.md
```diff
@@ -114,3 +114,18 @@ A setter method `setRoutePrefix` can be used to define a route prefix used to ma
...
+```yaml
+services:
+    ezsystems.platformui.navigationhub.link.admin.dashboard:
+        # ...
+        tags:
+            - name: ezplatform.ui.link
+              priority: 100
+```
+
+Lower `priority` values will make links appear first, those with the same `priority` value will appear in an order of definition, higher values obviously later.
```
dpobel, line 129: hum, usually when you call something `priority` higher values come first. Is it done this way in others contexts ? if so I guess we can keep it otherwise, I would really prefer the order to be inverted.

dpobel (approved):
other than the 2 remarks, +1

docs/howto/add_navigation_hub_items.md
bdunogier: Yes, that is my main doubt with that. The standardized `priority` in Symfony is used in reverse order. Using that to order graphical elements feels really odd to me.

Doc note: i would avoid terms such as "obviously". They're very subjective, and don't add much.

Nattfarinn: Hmm, true. The more I think about `priority` the less I like it, and I agree it may be misleading. Maybe we should go with `order`, as you suggested? (If so, I'll update Dashboard PR as well)

bdunogier: `order` is okay, I think. Even though it is still a bit wrong: you "order" something, you don't "have an order", unless you're a shop. `index` or `position` are more correct, but they may hint for an "exact" position.

Also, now that I think about it, Location's priority is used in an ascending order... but we can say I think that they are not the same domains (one is services definitions, aimed at developers/integrators, and the other content, aimed at editors).

PR #95, opened 2017-07-28
[doc] How-to: Run the Universal Discovery

dpobel:
Documentation for https://jira.ez.no/browse/EZP-27657

PR #97, opened 2017-08-01
EZP-27670: Move a content item using action bar

yannickroger:
Link: https://jira.ez.no/browse/EZP-27670

## Description
Enables to move an item thanks to the Move button in the action bar.

## Tests
Manual tests and phpspec

### DIFF
--- a/src/bundle/Resources/views/locationview.html.twig
+++ b/src/bundle/Resources/views/locationview.html.twig
@@ -89,6 +89,29 @@
                     {{ 'actionbar.trash'|trans|desc('Send to Trash')}}
                 </button>
             {{ form_end(trashLocationForm) }}
+
+            {{
+                form_start(moveLocationForm, {
+                    "action": path("ez_hybrid_platform_ui_location_move_location",
+                    {'contentId': content.id, 'locationId': locationId})
+                })
+            }}
+                {{ form_widget(moveLocationForm.newParentLocationId) }}
+                {{
+                    form_widget(moveLocationForm.move, {
+                        'label': 'actionbar.move'|trans()|desc('Move'),
+                        'attr': {
+                            'class': 'ez-button ez-button-action ez-js-run-universal-discovery',
+                            'data-ud-title': 'actionbar.move.selectbox.title'|trans()|desc('Select a new location to move your content item'),
+                            'data-ud-confirm-label': 'actionbar.move.selectbox.confirmlabel'|trans()|desc('Move'),
+                            'data-ud-starting-location-id': location.parentLocationId,
+                            'data-ud-container': true,
+                            'data-ud-confirm-fill': '#' ~ moveLocationForm.newParentLocationId.vars.id,
+                            'data-ud-confirm-fill-with': 'location.id',
+                        }
+                    })
+                }}
+            {{ form_end(moveLocationForm) }}
         {% endblock %}
     </ez-toolbar>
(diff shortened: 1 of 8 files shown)


src/bundle/Resources/views/locationview.html.twig
```diff
@@ -89,6 +89,28 @@
...
+                })
+            }}
+                {{ form_widget(moveLocationForm.newParentLocationId) }}
+
+                {{
+                    form_widget(moveLocationForm.move, {
+                        'label': 'actionbar.move'|trans()|desc('Move'),
+                        'attr': {
+                            'class': 'ez-button ez-button-action ez-js-run-universal-discovery',
+                            'data-ud-title': 'actionbar.move.selectbox.title'|trans()|desc('Select the location where you want to move the location'),
```
SylvainGuittard: Title: 
`Select a new location to move your content item`

src/bundle/Resources/views/locationview.html.twig
```diff
@@ -89,6 +89,28 @@
...
+                            'data-ud-container': true,
+                            'data-ud-confirm-fill': '#' ~ moveLocationForm.newParentLocationId.vars.id,
+                            'data-ud-confirm-fill-with': 'location.id',
```
SylvainGuittard: Could you please add `'data-ud-confirm-label': 'Move'`?
reference: https://github.com/ezsystems/hybrid-platform-ui/pull/95/files#diff-89cf99eb2ca64e40720b4f0294ec5407R45

yannickroger:
@SylvainGuittard I took care of your feedback, also I added the disabled for root location as asked in the issue.

src/lib/Repository/UiLocationService.php
```diff
@@ -115,6 +115,18 @@ public function trashLocationAndReturnParent(Location $location)
...
     /**
+     * Checks if a given location is root location.
+     *
+     * @param Location $location
+     *
+     * @return bool true if the location is root
+     */
+    private function isRootLocation(Location $location)
+    {
+        return $location->contentId == 1;
```
dpobel: In v1, Locations with a depth of 1 ie `/`, `/Media`, `/Users` (and `/Setup`) can not be moved as well as the *virtual root Location*, so I guess, the same should be implemented meaning you should probably test against the Location depth.

src/lib/Repository/UiLocationService.php
```diff
@@ -123,7 +135,7 @@ public function trashLocationAndReturnParent(Location $location)
      */
     public function canRemoveLocation(Location $location)
     {
-        if ($location->contentId == 1) {
+        if ($this->isRootLocation($location)) {
```
dpobel, line 138: you could just write:
```php
return !$this->isRootLocation($location);
```

dpobel: and same in `canMoveLocation`

dpobel (approved):
besides the logic remark around *root* Location(s), +1

src/lib/Repository/UiLocationService.php
yannickroger: but how will it do the permission check that comes afterwards?

dpobel: ah yeah missed this one, so this comment only applies to `canMoveLocation` :)

PR #99, opened 2017-08-18
EZP-27696: Hybrid Platform UI Dashboard blocks

Nattfarinn:
**Jira ticket:** https://jira.ez.no/browse/EZP-27696
----

Depends on: https://github.com/ezsystems/hybrid-platform-ui/pull/104
----
### All Content: https://github.com/ezsystems/hybrid-platform-ui/pull/99/commits/85d4b46e518129a023491d5daa3e189f6085231c

Implements:

- [EZP-27704](https://jira.ez.no/browse/EZP-27704) - As an Editor, I want to see a list of all content items which are created/owned by me in a widget labelled ‘Me’ split in 3 tabs - My drafts, My content, and My media
- [EZP-27705](https://jira.ez.no/browse/EZP-27705) - As an Editor, I want to see a list of all content items in the repository in a widget labelled ‘Everyone’ split in 2 tabs - Content and Media
- [EZP-27706](https://jira.ez.no/browse/EZP-27706) - As a Developer, I want to load content asynchronously for each tab
- [EZP-27707](https://jira.ez.no/browse/EZP-27707) - As an Editor, I want to have the option to edit a draft
- [EZP-27708](https://jira.ez.no/browse/EZP-27708) - As an Editor, I want to have the option to edit and view a published content item
- [EZP-27709](https://jira.ez.no/browse/EZP-27709) - As an Editor, I will see pagination option when the number of content items exceed the allowed limit of 10 in the table
- [EZP-27710](https://jira.ez.no/browse/EZP-27710) - Implement the new UI guidelines for Tabs and table

### DIFF
--- a/src/bundle/Controller/DashboardController.php
+++ b/src/bundle/Controller/DashboardController.php
@@ -9,14 +9,38 @@
 namespace EzSystems\HybridPlatformUiBundle\Controller;
 
 use EzSystems\HybridPlatformUi\Components\MainContent;
+use EzSystems\HybridPlatformUi\Dashboard\Dashboard;
 use EzSystems\PlatformUIBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
 
 class DashboardController extends Controller
 {
+    protected $dashboard;
+
+    public function __construct(Dashboard $dashboard)
+    {
+        $this->dashboard = $dashboard;
+    }
+
     public function viewDashboardAction(MainContent $mainContent)
     {
         $mainContent->setTemplate('EzSystemsHybridPlatformUiBundle::dashboard.html.twig');
+        $mainContent->setParameters(['dashboard' => $this->dashboard]);
 
         return $mainContent;
+    }
+
+    public function viewTabAction(Request $request, $sectionIdentifier, $tabIdentifier)
+    {
+        $content = $this->dashboard
+            ->getSection($sectionIdentifier)
+            ->getTab($tabIdentifier);
+
+        return new Response(
+            $content->render(
+                $request->get('page', 1)
+            )
+        );
     }
 }
--- a/src/bundle/Resources/config/dashboard.yml
+++ b/src/bundle/Resources/config/dashboard.yml
@@ -6,3 +6,49 @@ services:
 
     EzSystems\HybridPlatformUi\Dashboard\:
         resource: '../../../lib/Dashboard/*'
+
+    EzSystems\HybridPlatformUi\Dashboard\Section\MyContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_section.html.twig']]
+            - [setParameter, ['identifier', 'my_content']]
+        tags:
+            - name: ezplatform.ui.dashboard.section
+              identifier: my_content
+              order: 1
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\MyDrafts:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_drafts_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'my_content']]
+            - [setParameter, ['identifier', 'my_drafts']]
+        tags:
+            - name: ezplatform.ui.dashboard.tab
+              section: my_content
+              identifier: my_drafts
+              order: 1
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\MyContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'my_content']]
+            - [setParameter, ['identifier', 'my_content']]
+        arguments:
+            $permissionResolver: '@=service("ezpublish.api.repository").getPermissionResolver()'
+        tags:
+            - name: ezplatform.ui.dashboard.tab
+              section: my_content
+              identifier: my_content
+              order: 10
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\MyMedia:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'my_content']]
+            - [setParameter, ['identifier', 'my_media']]
+        arguments:
+            $permissionResolver: '@=service("ezpublish.api.repository").getPermissionResolver()'
+        tags:
+            - name: ezplatform.ui.dashboard.tab
+              section: my_content
+              identifier: my_media
+              order: 20
--- a/src/bundle/Resources/config/dashboard.yml
+++ b/src/bundle/Resources/config/dashboard.yml
@@ -52,3 +52,34 @@ services:
               section: my_content
               identifier: my_media
               order: 20
+
+    EzSystems\HybridPlatformUi\Dashboard\Section\AllContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:all_content_section.html.twig']]
+            - [setParameter, ['identifier', 'all_content']]
+        tags:
+            - name: ezplatform.ui.dashboard.section
+              identifier: all_content
+              order: 10
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\AllContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:all_content_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'all_content']]
+            - [setParameter, ['identifier', 'all_content']]
+        tags:
+            - name: ezplatform.ui.dashboard.tab
+              section: all_content
+              identifier: all_content
+              order: 1
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\AllMedia:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:all_content_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'all_content']]
+            - [setParameter, ['identifier', 'all_media']]
+        tags:
+            - name: ezplatform.ui.dashboard.tab
+              section: all_content
+              identifier: all_media
+              order: 10
--- a/src/bundle/Resources/config/dashboard.yml
+++ b/src/bundle/Resources/config/dashboard.yml
@@ -1,0 +1,8 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    EzSystems\HybridPlatformUi\Dashboard\:
+        resource: '../../../lib/Dashboard/*'
--- a/src/bundle/Resources/public/css/modules/tabs.css
+++ b/src/bundle/Resources/public/css/modules/tabs.css
@@ -27,3 +27,26 @@
     background: var(--ez-color-ground-base);
     font-weight: bold;
 }
+
+.ez-dashboard-tabs {
+    background: var(--ez-color-ground-base);
+    width: 90%;
+    margin: 0 auto;
+    padding: 1em;
+    border-radius: 5px;
+}
+
+.ez-dashboard-tabs .ez-tabs-label.is-tab-selected a {
+    color: var(--ez-color-text-base-inverted);
+    background: var(--ez-color-ground-secondary);
+}
+
+.ez-dashboard-tabs h3 {
+    margin-top: 0;
+}
+
+.ez-dashboard-tabs .ez-tabs-list {
+    background: var(--ez-color-ground-base);
+    border-bottom: 5px solid var(--ez-color-ground-secondary);
+    margin-bottom: 1em;
+}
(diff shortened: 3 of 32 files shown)


src/bundle/Resources/config/dashboard.yml
```diff
@@ -0,0 +1,87 @@
+services:
+    _defaults:
+        autowire: true
+        autoconfigure: true
+        public: false
+
+    EzSystems\HybridPlatformUi\Dashboard\:
+        resource: '../../../lib/Dashboard/*'
+
+    EzSystems\HybridPlatformUi\Dashboard\Dashboard:
```
SylvainGuittard: Why is it empty? Normal?

src/bundle/Resources/config/dashboard.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    EzSystems\HybridPlatformUi\Dashboard\Section\MyContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_section.html.twig']]
+            - [setParameter, ['identifier', 'my_content']]
+        tags:
+            - name: ezplatform.ui.dashboard.section
+              identifier: my_content
+              position: 1
```
SylvainGuittard: How to use `position`? You have several tables with `position: 1`.

src/bundle/Resources/public/css/modules/tabs.css
```diff
@@ -27,3 +27,26 @@
...
+.ez-dashboard-tabs .ez-tabs-label.is-tab-selected a {
+    color: var(--ez-color-text-base-inverted);
+    background: var(--ez-color-ground-secondary);
+}
+
+.ez-dashboard-tabs h3 {
+    margin-top: 0;
+}
+
+.ez-dashboard-tabs  .ez-tabs-list {
```
SylvainGuittard: extra space between `.ez-dashboard-tabs` and `.ez-tabs-list`

src/bundle/Resources/config/dashboard.yml
Nattfarinn: Yes, it is. :) Our Dashboard class definition does not require anything else, but still have to be defined to be injectable by Dependency Injection layer.

src/bundle/Controller/DashboardController.php
```diff
@@ -9,14 +9,36 @@
...
+
     public function viewDashboardAction(MainContent $mainContent)
     {
         $mainContent->setTemplate('EzSystemsHybridPlatformUiBundle::dashboard.html.twig');
+        $mainContent->setParameters([ 'dashboard' => $this->dashboard ]);
 
         return $mainContent;
     }
+
+    public function viewTabAction(Request $request, $section, $tab)
```
andrerom: Argument names should probably be made more clear as `$sectionIdentifier, $tabIdentifier`, maybe also type hint on them

src/bundle/Resources/config/dashboard.yml
```diff
@@ -0,0 +1,87 @@
...
+              identifier: my_drafts
+              position: 1
+
+    EzSystems\HybridPlatformUi\Dashboard\Tab\MyContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_tab.html.twig']]
+            - [setParameter, ['section_identifier', 'my_content']]
+            - [setParameter, ['identifier', 'my_content']]
+        arguments:
+            $permissionResolver: '@=service("ezpublish.api.inner_repository").getPermissionResolver()'
```
andrerom: why inner repository?

same other places, that's kind of internal logic so ideally anything above should not be aware of it unless of special cases.

src/bundle/Resources/config/dashboard.yml
Nattfarinn: `position` name is something to discuss (maybe we should name it `priority` here?). It works like `priority` used in Symfony. Lower `position` will appear first, those with the same position will appear in an order of definition, higher obviously later.

Here, every element has it own position unique to his location:
- `1, 10, 20` for tabs in `my_content` section,
- `1, 10` for tabs in `all_content` section
- and `1, 10` to sections `my_content` and `all_content` (respectively).

src/bundle/Resources/config/dashboard.yml
Nattfarinn: Oops. You didn't see that.

src/bundle/Resources/config/dashboard.yml
SylvainGuittard: Gotcha!

andrerom: if #92 ends up with `priority`, maybe this should as well so it's clear what it does.

src/bundle/Resources/config/dashboard.yml
bdunogier: You don't need to define it in that case, since it is autoconfigured by the line above. It will be created automatically with the class name as the service's id.

src/bundle/Resources/translations/messages.en.yml
```diff
@@ -0,0 +1,18 @@
+dashboard.section.my_content: Me
```
bdunogier: Why did you go with yaml format ? We use xlf most of the time (the only exception at the moment is field groups names).

src/bundle/Resources/config/dashboard.yml
```diff
@@ -0,0 +1,87 @@
...
+
+    EzSystems\HybridPlatformUi\Dashboard\:
+        resource: '../../../lib/Dashboard/*'
+
+    EzSystems\HybridPlatformUi\Dashboard\Dashboard:
+
+    EzSystems\HybridPlatformUi\Dashboard\Section\MyContent:
+        calls:
+            - [setTemplate, ['EzSystemsHybridPlatformUiBundle:dashboard:my_content_section.html.twig']]
+            - [setParameter, ['identifier', 'my_content']]
```
bdunogier, line 13: Why do we need this line ? Does it not repeat `identifier: my_content` below ?
The same question goes for tabs definition below and `section_identifier`.

src/bundle/Resources/views/dashboard/all_content_tab.html.twig
```diff
@@ -0,0 +1,34 @@
+{% if data|length %}
+    <table class="ez-table-data">
+        <thead>
+        <tr>
+            <th>{{ 'dashboard.table.name'|trans }}</th>
```
bdunogier: We use the `desc` function from jms translation bundle to define the default translation, so that it gets dumped by the translation extractor (see other templates from the same package).

bdunogier (changes requested):
Having both the framework for dashboard blocks and all of the implementations in one pull-request makes the review more cumbersome.

Could you move the blocks, besides maybe an example one, out of this pull-request to a new one ?

Furthermore, we have so far managed to cover most of the classes with phpspec (except for controlles). This PR doesn't have any, and it bothers me a bit.

bdunogier:
Side note: the acronym we use is HPUI, not HBUI. As in Hybrid Platform :)
On the other hand, you don't really need to include it, as the PR is against hybrid-platform-ui.

Nattfarinn:
> Side note: the acronym we use is HPUI, not HBUI. As in Hybrid Platform :)

I copied title from Epic :)

bdunogier:
> Side note: the acronym we use is HPUI, not HBUI. As in Hybrid Platform :)

Then it is wrong in the epic, and it must be changed here as well.
Also, pull-requests don't have to be copies of the epic. First, we usually have more than one PR to implement an epic. Furthermore, the epic describes the expectation, while the pull-request implements something. In addition, the Epic is on an issue tracker that is about the whole software (hence the "HPUI" in the title), not just about a component (as opposed to this package).

As an example, "Dashboard framework with all blocks and pagination support" would fit better.

Nattfarinn:
@bdunogier I made separate PR with base dashboard abstraction: https://github.com/ezsystems/hybrid-platform-ui/pull/104

This branch also contains same code because it's directly related, but to make PR a bit simpler I've also moved code to two separate commits as well:
 - My Content block: https://github.com/ezsystems/hybrid-platform-ui/pull/99/commits/d2bc3c2154b63ee4346b7c6586ee072aeacc8ac0
 - All Content block: https://github.com/ezsystems/hybrid-platform-ui/pull/99/commits/85d4b46e518129a023491d5daa3e189f6085231c

src/bundle/Resources/views/dashboard/my_content_section.html.twig
```diff
@@ -0,0 +1,26 @@
+<section class='ez-tabs ez-dashboard-tabs'>
```
dpobel, line 1: it's almost the same template as all_content_section.html.twig, this leads me thinking that we should probably have a `TabBasedSection` class/interface somewhere

src/bundle/Resources/views/dashboard/my_content_tab.html.twig
```diff
@@ -0,0 +1,34 @@
...
+        </tbody>
+    </table>
+    {% if pages_urls|length > 1 %}
+        <ez-pagination
+                current-page="{{ current_page }}"
+                pages-urls='{{ pages_urls|json_encode(constant('JSON_FORCE_OBJECT')) }}'
+        ></ez-pagination>
+    {% endif %}
+{% else %}
+    <p class="ez-table-empty">{{ 'dashboard.tab.my_content.empty'|trans|desc('No content items. Content items you create will appear here') }}</p>
```
dpobel, line 33: according to screeshots in the EPIC, when there's no content to display, we should still display the table header

src/bundle/Resources/views/dashboard/my_content_tab.html.twig
```diff
@@ -0,0 +1,34 @@
...
+                        'contentId': row.contentId,
+                        'versionNo': row.version,
+                        'language': row.language
+                    }) }}" class="ez-button">{{ 'dashboard.button.edit'|trans|desc('Edit') }}</a></td>
+            </tr>
+        {% endfor %}
+        </tbody>
+    </table>
+    {% if pages_urls|length > 1 %}
+        <ez-pagination
```
dpobel, line 27: I don't know this one but I guess it is to implement a local to the tab navigation. That's not needed, as documented in https://github.com/ezsystems/hybrid-platform-ui/blob/master/docs/howto/render_tabs.md#links, you can use the class `ez-js-local-navigation` so that links acts on the `ez-asynchronous-block` instead of loading a full new page. So basically here, you can write something like:

```twig
<nav class="ez-pagination ez-js-local-navigation">
    <ul>
    {% for k, page in pages_urls %}
        <li><a href="{{ page }}">{{ k }}</a></li>
    {% endfor %}
    </ul>
</nav>
```

src/bundle/Resources/views/dashboard/my_content_tab.html.twig
```diff
@@ -0,0 +1,34 @@
...
+            <th>{{ 'dashboard.table.last_saved'|trans|desc('Last Saved') }}</th>
+            <th></th>
+        </tr>
+        </thead>
+        <tbody>
+        {% for row in data %}
+            <tr>
+                <td><a href="{{ url('_ez_content_view', { 'contentId': row.contentId }) }}">{{ row.name }}</a></td>
+                <td>{{ row.type }}</td>
+                <td>{{ row.modified|date('M d, Y h:iA') }}</td>
```
dpobel, line 16: you should use `localizeddate` instead

src/bundle/Resources/views/dashboard/my_content_section.html.twig
```diff
@@ -0,0 +1,26 @@
...
+        {% for tab_name, tab in tabs %}
+            <li class="ez-tabs-label{% if loop.first %} is-tab-selected{% endif %}">
+                {% set display_name = tab_name|split('_')|join(' ')|capitalize %}
+                <a href="#ez-dashboard-{{ identifier }}-tab-{{ tab_name }}">{{ ('dashboard.tab.' ~ tab_name)|trans|desc(display_name) }}</a>
+            </li>
+        {% endfor %}
+    </ul>
+
+    {% for tab_name, tab in tabs %}
+        <ez-asynchronous-block class="ez-tabs-panel{% if loop.first %} is-tab-selected{% endif %}" id="ez-dashboard-{{ identifier }}-tab-{{ tab_name }}" url="{{ url('ez_hybrid_platform_ui_dashboard_view_tab', {
```
dpobel, line 14: for the first asynchronous block (the one filled by default), you have to add the `loaded` attribute otherwise the tab content will be reloaded if you switch to another tab and then switch back to this one.

src/lib/Dashboard/Tab/AllContent.php
```diff
@@ -0,0 +1,96 @@
...
+        $query->filter = new Query\Criterion\Subtree('/1/2/');
+        $query->sortClauses = [new Query\SortClause\DateModified(Query::SORT_DESC)];
+
+        $pager = new Pagerfanta(
+            new ContentSearchAdapter($query, $this->searchService)
+        );
+        $pager->setMaxPerPage(10);
+        $pager->setCurrentPage($page);
+
+        $this->setParameter('current_page', $page);
```
dpobel, line 87: hum, I would directly pass the `$pager` as a parameter, normally it contains everything that is needed for rendering. If not, it's probably because you need a custom adapter

dpobel (changes requested):
(:warning: it's not a full review, I quickly looked at the code to better understand #104).

In addition, to the inline comment, the tab templates look very much the same. So as for section, I'm tempted to suggest that we could maybe have a `PaginatedTableTab` class even if this seems a bit more complicated in that case.

PR #100, opened 2017-08-18
EZP-27740: Trash view

yannickroger:
Link(s): 
- Trash view: https://jira.ez.no/browse/EZP-27740
- Empty trash: https://jira.ez.no/browse/EZP-27741
- Restore: https://jira.ez.no/browse/EZP-27742

## Description
Trash view allowing to display trash items, empty the trash and restore items to their original location or a given one.

Screencast: https://drive.google.com/file/d/0B5amepf6EkXbX0lBVFpab1o4dVE/view

## Tests
Manual tests and phpspec

## TODO
- [x] Cleanup and rebase

### DIFF
--- a/src/bundle/Controller/TrashController.php
+++ b/src/bundle/Controller/TrashController.php
@@ -1,0 +1,96 @@
+<?php
+
+/**
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+namespace EzSystems\HybridPlatformUiBundle\Controller;
+
+use EzSystems\HybridPlatformUi\Components\MainContent;
+use EzSystems\HybridPlatformUi\Form\UiFormFactory;
+use EzSystems\HybridPlatformUi\Repository\UiTrashService;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\Request;
+
+class TrashController extends Controller
+{
+    /** @var UiTrashService */
+    public $uiTrashService;
+
+    /** @var UiFormFactory */
+    public $formFactory;
+
+    public function __construct(
+        UiTrashService $uiTrashService,
+        UiFormFactory $formFactory)
+    {
+        $this->uiTrashService = $uiTrashService;
+        $this->formFactory = $formFactory;
+    }
+
+    public function viewTrashAction(MainContent $mainContent)
+    {
+        $uiTrashItems = [];
+        $nbTrashItems = $this->uiTrashService->countTrashItems();
+        $hasTrashItems = false;
+
+        if ($nbTrashItems > 0) {
+            $uiTrashItems = $this->uiTrashService->getTrashItems();
+            $hasTrashItems = true;
+        }
+
+        $emptyTrashForm = $this->formFactory->createEmptyTrashForm(!$hasTrashItems);
+
+        $restoreTrashItemsForm = $this->formFactory->createRestoreTrashItemsForm(
+            $uiTrashItems
+        );
+
+        $mainContent->setTemplate('EzSystemsHybridPlatformUiBundle::trash.html.twig');
+        $mainContent->setParameters([
+            'nbTrashItems' => $nbTrashItems,
+            'uiTrashItems' => $uiTrashItems,
+            'emptyTrashForm' => $emptyTrashForm->createView(),
+            'restoreForm' => $restoreTrashItemsForm->createView(),
+        ]);
+
+        return $mainContent;
+    }
+
+    public function emptyTrashAction(Request $request)
+    {
+        $emptyTrashForm = $this->formFactory->createEmptyTrashForm();
+        $emptyTrashForm->handleRequest($request);
+
+        if ($emptyTrashForm->isValid()) {
+            $this->uiTrashService->emptyTrash();
+        }
+
+        return $this->refreshTrash();
+    }
+
+    public function restoreTrashItemsAction(Request $request)
+    {
+        $restoreTrashItemsForm = $this->formFactory->createRestoreTrashItemsForm();
+        $restoreTrashItemsForm->handleRequest($request);
+
+        if ($restoreTrashItemsForm->isValid()) {
+            $trashItemsIds = array_keys($restoreTrashItemsForm->get('trashItems')->getData());
+
+            if ($restoreTrashItemsForm->get('restore')->isClicked()) {
+                $this->uiTrashService->restoreTrashItems($trashItemsIds);
+            }
+
+            if ($restoreTrashItemsForm->get('restoreNewParent')->isClicked()) {
+                $parentLocationId = $restoreTrashItemsForm->get('newParentLocationId')->getData();
+                $this->uiTrashService->restoreTrashItems($trashItemsIds, $parentLocationId);
+            }
+        }
+
+        return $this->refreshTrash();
+    }
+
+    private function refreshTrash()
+    {
+        return $this->redirectToRoute('ez_hybrid_platform_ui_trash_view');
+    }
+}
--- a/src/bundle/Resources/config/routing.yml
+++ b/src/bundle/Resources/config/routing.yml
@@ -49,3 +49,18 @@ ez_hybrid_platform_ui_translation_actions:
     defaults:
         _controller: EzSystemsHybridPlatformUiBundle:Translation:actions
     methods: [POST]
+ez_hybrid_platform_ui_trash_view:
+    path: /trash
+    defaults:
+        _controller: EzSystemsHybridPlatformUiBundle:Trash:viewTrash
+    methods: [GET]
+ez_hybrid_platform_ui_empty_trash:
+    path: /trash/empty
+    defaults:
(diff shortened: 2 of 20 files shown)


src/bundle/Controller/TrashController.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        $this->uiTrashService = $uiTrashService;
+        $this->formFactory = $formFactory;
+    }
+
+    public function viewTrashAction(MainContent $mainContent)
+    {
+        $uiTrashItems = [];
+        $nbTrashItems = $this->uiTrashService->countTrashItems();
+        $isTrashItems = false;
```
dpobel: `$isTrashItems` ? `$hasTrashItems` maybe ?

src/bundle/Resources/config/routing.yml
```diff
@@ -49,3 +49,20 @@ ez_hybrid_platform_ui_translation_actions:
...
+    defaults:
+        _controller: EzSystemsHybridPlatformUiBundle:Trash:emptyTrash
+    methods: [POST]
+ez_hybrid_platform_ui_trash_restore:
+    path: /trash/restore
+    defaults:
+        _controller: EzSystemsHybridPlatformUiBundle:Trash:restoreTrashItems
+    methods: [POST]
+
+
```
dpobel: nitpick: useless empty lines

src/lib/Repository/Values/Content/UiTrashItem.php
```diff
@@ -0,0 +1,80 @@
...
+use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\API\Repository\Values\Content\TrashItem;
+use eZ\Publish\API\Repository\Values\ContentType\ContentType;
+
+/**
+ * Class UiTrashItem.
+ *
+ * Provides information needed to display a TrashItem including the TrashItem itself.
+ */
+class UiTrashItem
```
dpobel: I think this one should extend API TrashItem. It should also use ValueObject system for added properties.

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
+{% trans_default_domain 'trashview' %}
+
+<ez-server-side-content class="ez-trash-server-side-content">
+    {% block trashview %}
+    <section class="ez-trash-view">
```
dpobel: this element does not seem very useful, I think you can:
* remove it
* put the `ez-trash-view` class on `ez-server-side-content` element
* remove the class `ez-trash-server-side-content`

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
+{% trans_default_domain 'trashview' %}
```
dpobel: it would be nice to reuse the same context than in PlatformUI v1 and when possible the same keys

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
...
+            {{ form_widget(restoreForm.newParentLocationId) }}
+            {{ form_widget(restoreForm._token) }}
+
+            <div class="ez-list-toolbar">
+                <h2 class="ez-list-toolbar-label">{{ 'trash.buttons.title'|trans()|desc('Trash') }}</h2>
+                {% if uiTrashItems %}
+                    <div class="ez-list-toolbar-tools">
+                        {{
+                            form_widget(restoreForm.restoreNewParent, {
+                                'label': 'trash.buttons.restoreNewParent'|trans()|desc('Restore under new parent'),
```
dpobel: translate keys are dot separated also given we use a context, for new keys, no need to prefix strings, the key here should be `restore.under.new.parent` or something like that. This applies to all new keys in that template

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
...
+                                'attr': {
+                                    'class': 'ez-button ez-button-secondary ez-js-run-universal-discovery ez-js-activable-element',
+                                    'data-ud-title': 'trash.selectParent.title'|trans()|desc('Select the location where you want to restore the location'),
+                                    'data-ud-confirm-label': 'trash.selectParent.confirmlabel'|trans()|desc('Restore'),
+                                    'data-ud-container': true,
+                                    'data-ud-starting-location-id': 2,
```
dpobel: I think you should omit that parameter

src/bundle/Resources/public/css/modules/trash.css
```diff
@@ -0,0 +1,33 @@
...
+
+.ez-trash-view .ez-action-bar {
+    flex: 0 0 var(--ez-toolbar-width);
+    background: var(--ez-color-action-bar);
+    color: var(--ez-color-text-base-inverted);
+}
+
+.ez-trash-view .ez-trash-ancestor-trash {
+    font-style: italic;
+    opacity: 0.6;
```
dpobel: applying some `opacity` on text is a bit weird.
Most likely you need to use one of the colors defined in the UI guidelines https://ezsystems.github.io/ui-guidelines/#resources-colors (unsure if it is _Text base pale_ or _Text base light_ though) and reflected in custom properties in https://github.com/ezsystems/hybrid-platform-ui/blob/master/src/bundle/Resources/public/css/variables.css#L6

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
...
+                                {% else %}
+                                    {{ include('@EzSystemsHybridPlatformUi/fields/location/path.html.twig', {'locations': item.path}) }}
+                                {% endif %}
+                            </td>
+                        </tr>
+                    {% endfor %}
+                {% else %}
+                    <tr>
+                        <td colspan="4" class="ez-trash-empty">
+                            <span class="ez-trash-empty-title">
```
dpobel: `title` is kinda semantically wrong, either it's a title and you should use a `hN` tag, either it's not and then `title` should not be used :) anyway, it seems like you only want to a text in bold so a plain simple `<strong>` tag without any class would do perfectly the trick

src/bundle/Controller/TrashController.php
```diff
@@ -0,0 +1,96 @@
...
+
+        $emptyTrashForm = $this->formFactory->createEmptyTrashForm(!$isTrashItems);
+
+        $restoreTrashItemsForm = $this->formFactory->createRestoreTrashItemsForm(
+            $uiTrashItems
+        );
+
+        $mainContent->setTemplate('EzSystemsHybridPlatformUiBundle::trash.html.twig');
+        $mainContent->setParameters([
+            'nbTrashItems' => $nbTrashItems,
```
dpobel, line 50: if I'm not mistaken it's not used anywhere, so can be removed

src/bundle/Resources/public/css/modules/trash.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.ez-trash-view .ez-trash-empty {
+    font-style: italic;
+    opacity: 0.6;
```
dpobel: and same here

src/bundle/Resources/views/components/trash.html.twig
```diff
@@ -0,0 +1,3 @@
+{% trans_default_domain 'actionbar' %}
```
dpobel: same on context and keys, please try to reuse v1 contexts and keys

src/bundle/Resources/views/trash.html.twig
```diff
@@ -0,0 +1,111 @@
...
+                                    'disabled': 'disabled,'
+                                }
+                            })
+                        }}
+
+                        {{
+                            form_widget(restoreForm.restore, {
+                                'label': 'trash.buttons.restore'|trans()|desc('Restore'),
+                                'attr': {
+                                    'class': 'ez-button ez-button-negative ez-js-activable-element',
```
dpobel: it's not a _negative_ button but a primary one

src/bundle/Resources/public/css/modules/trash.css
```diff
@@ -0,0 +1,33 @@
...
+}
+
+.ez-trash-view .ez-trash-view-main {
+    padding: 1em 2em 1em 2em;
+    flex: 1 0;
+}
+
+.ez-trash-view .ez-action-bar {
+    flex: 0 0 var(--ez-toolbar-width);
+    background: var(--ez-color-action-bar);
```
dpobel: there are the exact same styles in content-view.css. While it's ok for the `flex` one since it kind of depends on `.ez-trash-view` or `ez-content-view`, for the background and color, this should be avoided, so it's time to create an `action-bar.css` that defines those

dpobel (changes requested):

src/bundle/Controller/TrashController.php
yannickroger: It is used here https://github.com/ezsystems/hybrid-platform-ui/pull/100/files#diff-111156c26282d4ecd9213484071ec6b6R92

Maybe I should make the line shorter.

dpobel: or I should open my eyes or start using Ctrl-F ;-)

dpobel (approved):

andrerom (approved):

PR #104, opened 2017-08-29
EZP-27696: Dashboard framework with pagination support

Nattfarinn:
Base PR for: https://github.com/ezsystems/hybrid-platform-ui/pull/99
----

To make Code Review a bit simpler, I moved framework part of epic to this separate PR. It contains only abstraction and base Dashboard implementation.

### DIFF
--- a/src/bundle/DependencyInjection/EzSystemsHybridPlatformUiExtension.php
+++ b/src/bundle/DependencyInjection/EzSystemsHybridPlatformUiExtension.php
@@ -30,6 +30,7 @@ class EzSystemsHybridPlatformUiExtension extends Extension implements PrependExt
         $loader->load('components.yml');
         $loader->load('pjax.yml');
         $loader->load('repository.yml');
+        $loader->load('dashboard.yml');
     }
 
     public function prepend(ContainerBuilder $container)
@@ -37,6 +38,7 @@ class EzSystemsHybridPlatformUiExtension extends Extension implements PrependExt
         $this->prependViewsConfiguration($container);
         $this->prependFosJsRoutingConfiguration($container);
         $this->prependLayoutConfiguration($container);
+        $this->prependRouterConfiguration($container);
     }
 
     private function prependViewsConfiguration(ContainerBuilder $container)
@@ -60,5 +62,20 @@ class EzSystemsHybridPlatformUiExtension extends Extension implements PrependExt
         $config = Yaml::parse(file_get_contents($configFile));
         $container->prependExtensionConfig('ezpublish', $config);
         $container->addResource(new FileResource($configFile));
+    }
+
+    private function prependRouterConfiguration(ContainerBuilder $container)
+    {
+        $config = [
+            'router' => [
+                'default_router' => [
+                    'non_siteaccess_aware_routes' => [
+                        'ez_hybrid_platform_ui_dashboard_view_tab',
+                    ],
+                ],
+            ],
+        ];
+
+        $container->prependExtensionConfig('ezpublish', $config);
     }
 }
--- a/src/bundle/Resources/config/navigationhub.yml
+++ b/src/bundle/Resources/config/navigationhub.yml
@@ -65,7 +65,7 @@ services:
         class: "%ezsystems.platformui.navigationhub.link.route.class%"
         arguments:
             - "@router"
-            - "admin_dashboard"
+            - "ez_hybrid_platform_ui_dashboard"
             - "Dashboard"
             - "admin"
         tags:
--- a/src/bundle/Resources/public/css/modules/table-data.css
+++ b/src/bundle/Resources/public/css/modules/table-data.css
@@ -40,3 +40,8 @@
     /* so that, if there's only one td (Role pages), it's left aligned */
     text-align: left;
 }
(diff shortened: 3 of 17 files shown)


src/bundle/DependencyInjection/EzSystemsHybridPlatformUiExtension.php
```diff
@@ -61,4 +63,19 @@ private function prependLayoutConfiguration(ContainerBuilder $container)
         $container->prependExtensionConfig('ezpublish', $config);
         $container->addResource(new FileResource($configFile));
     }
+
+    private function prependRouterConfiguration(ContainerBuilder $container)
+    {
+        $config = [
+            'router' => [
+                'default_router' => [
+                    'non_siteaccess_aware_routes' => [
```
dpobel, line 72: why do you make this route not siteaccess aware ?

src/bundle/Resources/config/navigationhub.yml
```diff
@@ -65,7 +65,7 @@ services:
         class: "%ezsystems.platformui.navigationhub.link.route.class%"
         arguments:
             - "@router"
-            - "admin_dashboard"
+            - "ez_hybrid_platform_ui_dashboard"
```
dpobel, line 68: hum, I think there's a confusion here. This changes basically replaces the *admin* Dashboard (the one accessible in Admin / Dashboard)  by the dashboard you are implementing here which is supposed to be visible right after login or by clicking on the eZ logo in the top left corner of the UI. Unless, this was asked by PM (and I failed to find anything related to that), we should keep the admin dashboard

src/bundle/Resources/public/css/modules/table-data.css
```diff
@@ -40,3 +40,8 @@
     /* so that, if there's only one td (Role pages), it's left aligned */
     text-align: left;
 }
+
+.ez-dashboard-tabs .ez-table-data thead tr {
```
dpobel, line 44: the CSS should not be in that PR, especially because it refers to some markup which is not added yet

src/bundle/Resources/views/dashboard.html.twig
```diff
@@ -1,3 +1,5 @@
 <ez-server-side-content>
-    <h1>Dashboard</h1>
```
dpobel, line 2: according to the screenshot in the EPIC, you should actually replace that with something like:
```html
    <h1>{{ 'my.dashboard'|trans({}, 'dashboard')|desc('My Dashboard') }}</h1>
</header>
```

src/lib/Dashboard/PaginatedTab.php
```diff
@@ -0,0 +1,118 @@
...
+            'page' => $page,
+        ]);
+    }
+
+    /**
+     * @param int $page
+     *
+     * @return string
+     */
+    public function render($page = 1): string
```
dpobel, line 114: the `$page` is not used and is not defined the TabInterface ?

src/lib/Dashboard/PaginatedTab.php
```diff
@@ -0,0 +1,118 @@
...
+        return $this->parameters[$parameter];
+    }
+
+    /**
+     * @param int $current
+     * @param int $last
+     *
+     * @return array
+     */
+    protected function getPaginatorUrls(int $current, int $last): array
```
dpobel, line 80: With `getPageUrl`, this does not seem to be used anywhere yet, so better to not include that just yet (also I find the implementation a bit weird but it's hard to understand everything without an example of use ;))

dpobel (changes requested):
in addition to inline comments, it would be nice to unit test the Dashboard class

src/lib/Dashboard/Dashboard.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Dashboard;
+
+class Dashboard
```
glye, line 5: Unsure, but this whole class is quite basic and looks like something we or even Symfony ought to have standard code for. Wondering if `ValueObject` or somesuch could be used...? Same for `PaginatedTab` and `TabsSection`.

src/lib/Dashboard/SectionInterface.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Dashboard;
+
+interface SectionInterface
```
glye, line 5: This is the same as `TabInterface`. They are small, but could we reduce duplication by having a common top level interface, or a `Renderable` trait, maybe?

src/lib/Dashboard/TabInterface.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace EzSystems\HybridPlatformUi\Dashboard;
+
+interface TabInterface
```
glye, line 5: This is the same as `SectionInterface`. They are small, but could we reduce duplication by having a common top level interface, or a `Renderable` trait, maybe?
Also, missing method doc.

glye (commented):
(Some notes added, but I'm uncertain about them, so not making this a change request)

PR #105, opened 2017-08-30
[Doc] Fix subtree link 4th param being array not int

andrerom:
Does anyone knows if the intention was to change implementation to actually take int here?

Or should it be array as this fixes doc for to allow for additional params in the future?


andrerom:
Actually, the doc further down made it clear it is intended to be array.

So fixed this and some other doc issues compared to implementation in fa1f6b33cfa9b3a0b4f000e39d3c5f90055e6a17

PR #106, opened 2017-09-01
[WIP] EZP-27407: Setup the icon system

dpobel:
JIRA: https://jira.ez.no/browse/EZP-27407

# Description

**WIP**: That's a very early wip prototype.

## Base idea/goal

To have icons rendered in the interface, the idea is to use SVG instead of an icon font mainly for the reasons exposed in [Seriously, Don’t Use Icon Fonts](https://cloudfour.com/thinks/seriously-dont-use-icon-fonts/). So for instance, to have a trash icon in a delete button, the markup should be something like:

```html
<button class="ez-button ez-button-negative">
  <svg class="ez-icon ez-icon-trash">
    <use xlink:href="/path/to/icons-sprite.svg#trash"></use>
  </svg>
  Delete
</button>
```

When writing this, this gives the following result: 
delete

The idea is also to use the same system (and sprite) for Content Type icons. Content Type icons are very similar to regular UI icons but have 2 main differences:

1. the icon can be chosen based on the Content Type identifier but since Content Types are dynamic, a fallback icon has to be used if no dedicated one can be found
1. it's very common to add custom icons for custom Content Types. So this use case has to be possible.

## Tasks

* [x] Test the SVG based markup in other context than buttons
* [x] Add a `ez_ui_icon` Twig function or a `ui_icon.html.twig` to include or a macro or ... ?
* [ ] Add a `ez_content_type_icon` Twig function (to apply the fallback)
* [ ] Add a script to build the icon sprite (it should add aliases for instance `#blogpost` and `#blog_post` should resolve in the same icon)
* [ ] _iconify_ all the things (tm)
  * [x] Add an `icon` parameter to Navigation Zone
  * [x] Convert Search and Browse components to use a template
  * [x] Handle icons on Symfony Form elements (form theme ?)
  * [ ] PlatformUIBundle parts (https://github.com/ezsystems/PlatformUIBundle/pull/897)
* [x] Fix the bug with icon in zone in the Navigation Hub (https://github.com/ezsystems/hybrid-platform-ui-core-components/pull/40)


dpobel:
self note: command line used to create the sprite

```
$ ./node_modules/.bin/svg-sprite -s --symbol-dest=src/bundle/Resources/public/img/ --ss="icons.svg"  src/bundle/Resources/public/img/icons/*.svg
```

dpobel:
Test for icons in different context: icons-context Some minor CSS changes are needed.

dpobel:
(Moved icons and sprite to separate branch to ease the review)

